PR #55, opened 2024-01-25
fix: correct a couple add repo error messages

johnmaguire:


garethgeorge (approved):

garethgeorge:
Looks great! Thanks for the typo fixes. Hadn't noticed the missing star before, looks like I forgot to mark the field required. I'll bundle up a fix for that in my next commit (or if you want to include it the change is that the outermost form element should be:

```
<Form.Item label="Password" required={true}>
```

Glad to hear it's working well for you and especially on the restore view! I felt that being able to restore specific directories is a common want.

johnmaguire:
@garethgeorge I was hoping it would be that simple but then I noticed the label is set slightly differently here than other forms:

https://github.com/garethgeorge/backrest/blob/ee6134af76c3e90f542f67b89b2571f060db5590/webui/src/views/AddRepoModal.tsx#L237-L238

versus, e.g.

https://github.com/garethgeorge/backrest/blob/ee6134af76c3e90f542f67b89b2571f060db5590/webui/src/views/AddRepoModal.tsx#L171-L181

And on the other required fields, I didn't see a `required` flag for `Form.Item` - it looked like it only existed in the `rules` section, which the password field also sets:

https://github.com/garethgeorge/backrest/blob/ee6134af76c3e90f542f67b89b2571f060db5590/webui/src/views/AddRepoModal.tsx#L241-L249

But as you can see, there is also a `Form.Item<Repo>` and `Form.Item` for the password, and I wasn't sure what the purpose of each was.

garethgeorge:
Right -- this implementation is a bit messy! I was learning the antd framework as I implemented this form (and it was the first that I added) so it's not as consistent as it could be. I'll clean some of this up and submit a fix in my next big PR.

In that case, it's valid to wrap a Form.Item in another Form.Item for layout purposes in antd, but there are some mistakes there w.r.t. the properties I have set. I don't think the outerone should be specifying name="password" as well as the inner one.

Issue #56, opened 2024-01-25
Schedule-less plans / on-demand backup of repos?


johnmaguire:
On my laptop, I prefer to trigger backups on-demand rather than using a schedule. It would be great if it were possible to create a plan without a schedule - or offer on-demand backups directly from the repo. The only issue I see with the latter is that I do still want to prune after a backup, and retention polices are currently configured on the plan.

By the way, how do retention polices work with multiple plans? Does the `restic forget` command target the plan tag so it only applies to snapshots created by a given plan? Is there a way to see the actual command that was run (or params passed to a restic library) in the operation list?

garethgeorge:
Sounds very reasonable, at the moment any valid cronexpr should be admissible as a schedule. The UI doesn't permit specifying a year (or some of the more interesting supported options) but by manually editing the cronexpr you can try setting it to a date far in the future to effectively disable backups.

I'll leave this open to track UI support for "no schedule" repos where the cronexpr will be left empty / no operations will be performed automatically.

There's also a valid argument for supporting the same for forget and prune operations (e.g. disabling maintenance operations).

ovizii:
Could we add an option to “pause” a plan? That way, it could have a schedule but be paused and still be available for manually running.

garethgeorge:
Added in latest release, implemented a "Disable Scheduling" option in the plan view which will disable running scheduled tasks for that plan. Operations can still be triggered manually from the Web UI.

johnmaguire:
Thank you @garethgeorge! Really happy to see this. If you add the ability to Sponsor you to your profile, I'd love to toss you a tip!

Powerbless:
> Added in latest release, implemented a "Disable Scheduling" option in the plan view which will disable running scheduled tasks for that plan. Operations can still be triggered manually from the Web UI.

Is there a way to Trigger a Backup/Operation trough a cli command oder Powershell? This would bei nice!

garethgeorge:
Hey -- take a look at https://github.com/garethgeorge/backrest/discussions/275 which documents doing this with curl on linux. I suspect the powershell invocation will look fairly similar :)

PR #72, opened 2024-02-08
feat: allow plans to be added with no schedule

garethgeorge:
Address https://github.com/garethgeorge/backrest/issues/56


garethgeorge:
wound up taking a different implementation approach here.

PR #90, opened 2024-02-14
feat: Add support for multiple sets of expected env vars per repo scheme

### LINKED ISSUE #83: Fix Azure blob repo config check
Right now, Backrest requires `AZURE_ACCOUNT_KEY` when configuring a Azure Blob Repo. However, according to the [documentation](https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html#microsoft-azure-blob-storage), both `AZURE_ACCOUNT_KEY` as well as `AZURE_ACCOUNT_SAS` are interchangeably valid.

dhilgarth:
Fixes #83

### DIFF
--- a/webui/src/views/AddRepoModal.tsx
+++ b/webui/src/views/AddRepoModal.tsx
@@ -451,11 +451,11 @@ export const AddRepoModal = ({
   );
 };
 
-const expectedEnvVars: { [scheme: string]: string[] } = {
-  s3: ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"],
-  b2: ["B2_ACCOUNT_ID", "B2_ACCOUNT_KEY"],
-  azure: ["AZURE_ACCOUNT_NAME", "AZURE_ACCOUNT_KEY"],
-  gs: ["GOOGLE_APPLICATION_CREDENTIALS", "GOOGLE_PROJECT_ID"],
+const expectedEnvVars: { [scheme: string]: string[][] } = {
+  s3: [["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]],
+  b2: [["B2_ACCOUNT_ID", "B2_ACCOUNT_KEY"]],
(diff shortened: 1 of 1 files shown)


dhilgarth:
I've changed to code to support multiple distinct sets of expected env vars per scheme, because for some schemes there is no overlap at all between the different possibilities, e.g. for Google.

I'm not entirely happy with the error message, but I think it's good for now.

garethgeorge:
Awesome, really like the approach you took. This closely matches how I was thinking about the fix.

Re: the error message formatting, I think this looks pretty solid. It looks somewhat long for azure but I'll take long and explicit over anything ambiguous :)

Thanks for contributing this!

garethgeorge:
Hi, Daniel -- I'm evaluating relicensing Backrest as AGPL-3.0 . Thanks again for contributing this pretty significant improvement to how Backrest validates env vars --

dhilgarth:
I'm generally against re-licensing, bad examples like Elasticsearch or Terraform come to mind. However, as I read the CLA, re-licensing to a fake OS license like the BSL should not be possible here, so that would be fine. I will take some time to read the CLA in detail in the next days. Please remind me if I don't get back to you by Monday.

garethgeorge:
Agreed and aware of those two examples :) main reason I'm looking at relicensing is that, relative to the base projects I'm building on (e.g. restic, etc) Backrest is becoming a very full featured program in a space where a lot of competing software is sold for money. I put a lot of time into Backrest and philosophically build it b/c I think it's right for there to be an open and fully _understandable_ (read: nonproprietary) solution for the average person to preserve their data so I want people building on it (even in SaaS contexts) to afford the same to their users. 

The SAAS loophole in GPL 3.0 makes me nervous re: other platforms using Backrest as a component of for-profit closed source storage as a service systems. 

The the CLA allows me to switch between offered by the free software foundation with the first step I have planned being a move to AGPL 3.0 which closes that loophole for closed source software that's accessed over a network.

PR #94, opened 2024-02-14
fix: Fix blank screen upon empty number input in Plan modal

### LINKED ISSUE #91: Issue: editing a plan retention policy may turn page blank
Hi Gareth,

this one is weird. I was editing a plan, and was changing from the default retention policy 30 count. I did the following steps: 

1. placed the cursor after the `30` value
2. pressed `Backspace`2 times to delete the number:
    - the 1st key press deletes the 0
    - the 2nd key press turns the full backrest page blank.

I then need to refresh the page for the content to appear again.

Placing the cursor before the number, and pressing the `DELETE` key twice has the same behavior.

I have the same behavior on Firefox and Edge.

dhilgarth:
fixes #91 

This is a pretty straight forward fix. I'm not sure if it isn't TOO straightforward and it should be handled differently

### DIFF
--- a/webui/src/views/AddPlanModal.tsx
+++ b/webui/src/views/AddPlanModal.tsx
@@ -1,3 +1,4 @@
+import { PartialMessage } from '@bufbuild/protobuf';
 import {
   Form,
   Modal,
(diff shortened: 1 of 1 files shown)


garethgeorge:
Ah, interesting. In general the way forms are handled right now is a bit of a mess, a big piece of the work on my plate right now is cleaning this up for the next release.

I think the right fix may actually be to just stop converting to a `new Plan` e.g.
```
<Typography>
     <pre>{JSON.stringify(form.getFieldsValue(), null, 2)}</pre>
</Typography>
```

Ostensibly this form's fields are supposed to always match the JSON structure of the plan proto and any validation errors will still be caught when trying to submit the form.

dhilgarth:
Yeah, I saw the difference to the Repo modal. I thought it was intentional, because the `JSON.stringify` outputs more properties than the stringified `Plan`.

PR #121, opened 2024-02-23
docs: update README.md docker-compose example

edersong:
Added the hostname option to the compose.yaml as Discussion #111


garethgeorge:
Nice! Setting the hostname in the docker-compose looks great.

PR #128, opened 2024-02-24
chore: add devcontainer and vscode launch configurations

Janhouse:
I added VSCode Dev Container configuration, which adds effortless dev environment setup for those who use VSCode and Docker.

It includes VSCode launch configurations for backend and frontend.

From what I tested, all the core dependencies are installed during creation of the container, and npm packages are installed when launching project.

Maybe it could be improved slightly to wait for backend to start before starting frontend, but as I see it, it is not that important.

Here is a short video on how I start it in a few minutes.

[Screencast from 2024-02-24 03-33-37.webm](https://github.com/garethgeorge/backrest/assets/1036439/f1ac73f9-1047-4265-aa66-6baf36f30075)

I tested only on Linux but it should work on Windows and MacOS as well, of course it starts it in a Linux container. Still, I guess most people would run Backrest on a Linux server.


garethgeorge:
This is a really awesome contribution for people looking to get into making changes to the project, thanks for putting this together!

Issue #110, opened 2024-02-21
Basic http authentication for SSO


Janhouse:
Would be great if basic http authentication could be used to authenticate users.

With SSO systems like [Authentik you can use forward-auth](https://goauthentik.io/docs/providers/proxy/forward_auth) to [delegate authentication](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) to the SSO system.
It can then send custom headers to the application which can use those headers to authenticate as a specific user.

In forward-auth scenario either a custom `x-forwarded-username` header can be used (which is injected by the forward-auth mechanism), or a basic http authentication could be used (credentials are injected by the forward-auth mechannism).

Backrest seems to use bearer authentication which [would not work with Authentik](https://goauthentik.io/docs/providers/proxy/header_authentication#sending-authentication).
So either supporting basic auth or defining header with the username would be great addition to support SSO forward-auth.

garethgeorge:
Sgtm, this can be supported by backrest. Happy to accept a PR if that’s something you’re interested in. I probably lean a bit in the direction of adding a basic auth mode (as opposed to bearer) mode.

PR #129, opened 2024-02-24
feat: check for basic auth (#110)

Janhouse:
Added a simple check for HTTP Basic authorization. Seems to work fine with Authentik SSO passing the auth header.

This does not add `WWW-Authenticate` header since it normally expects to receive Bearer auth for JWT.

I am not familiar with go, so I presume that `bcrypt.CompareHashAndPassword` takes care of timing attacks and I don't have to add `ConstantTimeCompare` myself. Please correct me if I am wrong.

Maybe adding rate limiting for brute force detection would be a good idea. Should it be added only for Basic auth, or is it relevant for Bearer token as well?

### DIFF
--- a/internal/auth/middleware.go
+++ b/internal/auth/middleware.go
@@ -17,6 +17,17 @@ const UserContextKey contextKey = "user"
 
 func RequireAuthentication(h http.Handler, auth *Authenticator) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		username, password, usesBasicAuth := r.BasicAuth()
+		if usesBasicAuth {
+			user, err := auth.Login(username, password)
+			if err == nil {
+				ctx := context.WithValue(r.Context(), UserContextKey, user)
+				h.ServeHTTP(w, r.WithContext(ctx))
+				return
+			}
+		}
+
 		token, err := ParseBearerToken(r.Header.Get("Authorization"))
 		if err != nil {
 			http.Error(w, "Unauthorized (No Authorization Header)", http.StatusUnauthorized)


garethgeorge:
Looks great! Thanks for contributing this!

garethgeorge:
If you have a bit of bandwidth, a great followup to this would be a wiki page covering authentication if you want to add some details about how to use this with a common reverse proxy. I'd be happy to expand whatever you pull request with more details about the workings of the existing jwt based auth.

Janhouse:
Here is the guide:

# Basic authentication for SSO

HTTP basic authentication can be used together with [forward-auth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) to enable single-sign-on.

Forward-auth passes request headers to a 3rd party software, most commonly an SSO identity provider service, which can add additional headers to the proxied service (_Backrest_), in this case adding `Authorization: Basic ...`

All of the major web servers and proxies support forward-auth (Apache, Nginx, Traefik, Caddy, etc.), please consult their documentation on how to set it up.

## Example with Authentik

### Set username and password attributes
When using Authentik proxy provider for forward-auth, enable `Send HTTP-Basic Authentication` on the Backrest **provider**, set **attribute names** which will be read from either group or user profile.

image

### Bind group or user to Backrest application
In the Authentik applications section, navigate to your Backrest application, and go to `Policy/Group/User Binding`. Click `Bind existing policy` button.

Choose the appropriate Group or User to bind to the application and click `Create`.
image

### Set the username and password in group/user attributes

Add the previously chosen username and password attributes (with values from actual user in Backrest) in the group attributes section and click `Update`

image

Now when connecting to Backrest, Authentik will add HTTP-Basic authentication header with username and password from the group/user attributes and user should be automatically logged in.

garethgeorge:

Janhouse:
Sure thing, AGPL sounds great.

PR #138, opened 2024-02-28
fix: install rclone with apk for alpine image

ArjenR:
Alpine is very much up to date with rclone. Using their provided rclone will probably be more stable due for lack of library and dependency issues.


garethgeorge:
sgtm, this seems reasonable to me & will probably do a better job of keeping backrest up to date with stable alpine releases.

PR #155, opened 2024-03-18
feat: automatically remove Apples quarantine flag

MarvinJWendt:
Apple usually puts a quarantine flag on downloaded binaries, that are not signed. This prevents execution of the file. This PR adds a command to the `install.sh` file, to automatically remove the quarantine flag of the installed binary.

*tested on macOS 14.x*

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -74,6 +74,7 @@ if [ "$OS" = "Darwin" ]; then
   install_unix
   create_launchd_plist
   enable_launchd_plist
+  sudo xattr -r -d com.apple.quarantine /usr/local/bin/backrest # remove quarantine flag
 elif [ "$OS" = "Linux" ]; then
   echo "Installing on Linux"
   install_unix


install.sh
```diff
@@ -74,6 +74,7 @@ if [ "$OS" = "Darwin" ]; then
   install_unix
   create_launchd_plist
   enable_launchd_plist
+  sudo xattr -r -d com.apple.quarantine /usr/local/bin/backrest # remove quarantine flag
```
garethgeorge, line 77: nit: let's pull off `-r` since it's operating directly on the binary! Shouldn't be needed.

garethgeorge:
This looks great, thanks for the PR!

install.sh
MarvinJWendt: True, I'll fix it when I am back home :)

garethgeorge: Thanks! As a side note I see you maintain https://github.com/installer/instl, how maintained is it / would it be a good candidate for download metrics for backrest?

MarvinJWendt: Instl is usable, but I don't actively work on it at the moment. I want to revive it, tho, as I still love the idea.
However, GitHub features a download count API, which can be utilized via shields.io do make a badge:

https://img.shields.io/github/downloads/garethgeorge/backrest/total

```
![GitHub Downloads (all assets, all releases)](https://img.shields.io/github/downloads/garethgeorge/backrest/total)
```
GitHub Downloads (all assets, all releases)

You can configure the badge with additional settings here:
https://shields.io/badges/git-hub-downloads-all-assets-all-releases

garethgeorge:

garethgeorge:
Haha, much appreciated :)

PR #187, opened 2024-04-02
feat: add a Bash script to help Linux user manage Backrest

Nebulosa-Cat:
This Bash script include those feature:

1. Install / Update / Uninstall Backrest
2. Start / Stop / Check Status of Backrest

When it install Backrest it will check CPU/Kernel tpye ( x86_64, Armv6, Arm64), if not those type it should stop the install process.

After check CPU/Kernel tpye, it will use `wget` to download the correct version of Backrest install pack from newest release. Then create a folder/path name "backrest" under `/home/username`, then unpackage the install pack the install pack and run installer.sh inside it, after that it will clean the install file (remove the `/home/username/backrest`).

Uninstall feature is same as install but excute the `uninstall.sh` script.

other feature use `sudo systemctl` command to manage it.

Which plateform this script should support:
Linux running on x86_64, armv6~v8 CPU/Kernel

I have test the x86_64 and Armv8, other version I don't have device or enviroment to test it.

### DIFF
--- a/scripts/get.sh
+++ b/scripts/get.sh
@@ -1,0 +1,148 @@
+#!/bin/bash
+PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin:~/bin
+export PATH
+
+#=================================================
+#	System Required: Linux
+#	Description: Backrest Helper
+#	Author: Nebulosa Cat
+#	WebSite: https://nebulosa-cat.moe
+#=================================================
+
+userHomeDir=$( getent passwd "$USER" | cut -d: -f6 )
+
+#Coler the Bash output
+Green_font_prefix="\033[32m" && Red_font_prefix="\033[31m" && Green_background_prefix="\033[42;37m" && Red_background_prefix="\033[41;37m" && Font_color_suffix="\033[0m" && Yellow_font_prefix="\033[0;33m"
+Info="${Green_font_prefix}[Info]${Font_color_suffix}"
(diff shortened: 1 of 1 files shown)


garethgeorge:
Hey! Thanks for sending this out, just pinging in to let you know this is on my plate -- will aim to merge this weekend after I've had some time to play with this, test on MacOS, possibly cover usage in the docs.

Nebulosa-Cat:
and if possible I think might be need to notice if user's restic, rclone is install as root or they need to backup root file (like /root or /var) maybe need use "sudo" when they run this script

when the user install not use `sudo get.sh` will create a systemd using `user` (user) and `user` (group)
config file will in `/home/user/.config/backrest`
cache file will in `/home/user/.local/share/backrest`, the `/home/user/.local/share/backrest/restic-0.16.4` or similar file's permission should be `rwxr-xr-x`, this must be check or the all backup function or index will error.
image

if use `sudo get.sh` to install this will install as root use
config file will in `/root/.config/backrest`
cache file will in `/root/.local/backrest`

and all restic or rclone config file should in same user (any user or root)

garethgeorge:
Hey thanks for contributing this! I just noticed after merging that you have a header re: attributions. I hope it's alright if I go ahead and update your website / author comment to a link to your github, I'm thinking this 

```
#=================================================
#	System Required: Linux
#	Description: Backrest Helper
#	Contributed-By: https://github.com/Nebulosa-Cat
#=================================================
```
I'm a bit more comfortable with that relative to an arbitrary domain -- hope that's alright :)

Otherwise looks awesome.

PR #191, opened 2024-04-04
fix(css): Fixing overflow issue

StasiumDev:
Fixing overflow issue when the screen is overflowing, causing the nav to overflow to right.

Preview:

https://github.com/garethgeorge/backrest/assets/41346624/1d2c837b-462b-4223-8a91-1067e2812090

It kept annoying me while trying out the tool, so I decided to look into it and fixed it.


garethgeorge:
Thanks for contributing this! Lgtm.

Issue #193, opened 2024-04-05
Feature Request: Homebrew tap


johnmaguire:
Note: if you have a question or want discussion please post in the [discussions area](https://github.com/garethgeorge/backrest/wiki).

**Is your feature request related to a problem? Please describe.**
Using an `install.sh` script doesn't feel very clean, as the installed files aren't tracked by my package manager. Additionally, when running it, I got an error during start, because I hadn't right click > Open'd the executable, to tell the system to trust it, after it was downloaded in the browser.

Shipping the binary with homebrew should avoid the need to right click > Open it on first run, and would allow me to update it with the rest of my packages.

**Describe the solution you'd like**
The ability to run `brew install backrest`

**Additional context**
Goreleaser has support for homebrew: https://goreleaser.com/customization/homebrew/

garethgeorge:
Thanks again for contributing this! Just pinging here to say the first brew tap is published with the `0.17.0` release https://github.com/garethgeorge/homebrew-backrest-tap/blob/main/backrest.rb !

PR #194, opened 2024-04-05
Add homebrew support

johnmaguire:
I took a stab at #193. It seems to work. You need a PAT with repo permissions, and a repository that starts with `homebrew-` (e.g. `homebrew-backrest-tap`.) That's pretty much it.

You can see the formula it created here: https://github.com/johnmaguire/homebrew-backrest-tap

You can install it with `brew install johnmaguire/backrest-tap/backrest`. To enable the service, run `brew service start backrest`. I tried running the service as root, but got the following error, and I'm out of time to look at this for now:

```
panic: couldn't determine home directory: $HOME is not defined

goroutine 1 [running]:
github.com/garethgeorge/backrest/internal/config.getHomeDir()
	/home/runner/work/backrest/backrest/internal/config/environment.go:80 +0x68
github.com/garethgeorge/backrest/internal/config.DataDir()
	/home/runner/work/backrest/backrest/internal/config/environment.go:54 +0x98
github.com/garethgeorge/backrest/internal/resticinstaller.FindOrInstallResticBinary()
	/home/runner/work/backrest/backrest/internal/resticinstaller/resticinstaller.go:243 +0x3fc
main.main()
	/home/runner/work/backrest/backrest/backrest.go:38 +0x74
```

Alternatively, we could try to submit a formula to Homebrew core. This could be nice, as then users could just `brew install backrest`. I don't know how long the approval process takes though, so it might be nice to have a tap in the meantime anyway. A core formula would need to build from source: https://docs.brew.sh/Acceptable-Formulae

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -146,6 +146,8 @@ brews:
 
     service: |
       run opt_bin/"backrest"
+      error_log_path var/"log/backrest.log"
+      log_path var/"log/backrest.log"
 
 changelog:
   sort: asc
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -132,7 +132,7 @@ brews:
   - name: backrest
     homepage: https://github.com/garethgeorge/backrest
 
-    url_template: "https://github.com/garethgeorge/backrest/releases/download/{{ .Tag }}/{{ .ArtifactName }}"
+    url_template: "https://github.com/johnmaguire/backrest/releases/download/{{ .Tag }}/{{ .ArtifactName }}"
 
     commit_author:
       name: goreleaserbot
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -144,30 +144,7 @@ brews:
       branch: main
       token: "{{ .Env.GH_BREW_TOKEN }}"
 
-    plist: |
-      <?xml version="1.0" encoding="UTF-8"?>
-      <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-      <plist version="1.0">
-      <dict>
-          <key>Label</key>
-          <string>com.backrest</string>
-          <key>ProgramArguments</key>
-          <array>
-          <string>/usr/local/bin/backrest</string>
-          </array>
-          <key>KeepAlive</key>
-          <true/>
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -128,6 +128,27 @@ docker_manifests:
...
+      name: goreleaserbot
+      email: bot@goreleaser.com
+
+    repository:
+      owner: garethgeorge
+      name: homebrew-backrest-tap
+      branch: main
+      token: "{{ .Env.GH_BREW_TOKEN }}"
+
+    service: |
```
garethgeorge, line 147: This is _really_ neat. Didn't know that homebrew had the capability to manage service execution, this is strictly better than the `install.sh` creating a launchd agent directly from XML.

garethgeorge:
Thanks for contributing this! A brew tap seems like a much improved way to distribute backrest for macOS (especially seeing that brew also has an opinionated approach to running background services?) I'll aim to learn a bit more and review over the weekend.

Re: distributing binaries, https://github.com/Homebrew/homebrew-cask is a tap that accepts binary contributions. I wonder if they'd take backrest?

johnmaguire:
Yes - I do think we could submit to homebrew-cask! I will say that as a user, I generally prefer regular homebrew-core to homebrew-cask (just because they are managed separately, and most of my tools come from core. Usually casks are proprietary software). I wasn't sure how you felt about distributing it through homebrew-core, since the installed artifact wouldn't necessarily exactly match the artifact from the Releases page.

I went with the tap approach mainly because it looked like it could be supported using goreleaser without too much work. I'm happy to take a look at putting together a Core or Cask formula together instead! I have never created one though, so I'm not sure what the update story looks like there.

garethgeorge:
Agree on the homebrew-core front! I'm not sure what'd be involved in getting a tool like backrest to compile there though given the stack needed for the UI. I think a dependency on golang + npm would probably do it -- but in terms of quick easy installation a cask might be a better / less fragile choice (but it should be relatively straight forward to build on MacOS).

It looks like https://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request#cask-related-pull-request covers keeping it up to date -- seems you're right that we'd need a new PR to keep it up to date for each version. 

In that case, let's take this PR and I'll setup a tap for backrest -- once I hit a `1.0.0` stable release (which I'm aiming for some time in the next 6 months) I'll contribute that to homebrew cask !

PR #195, opened 2024-04-05
Open docs links in a new tab

johnmaguire:
Some docs links already did this, this makes the rest of them match.

I was worried when I clicked on one of the links in a form tooltip and the page navigated away that I might've lost what I'd typed in.


garethgeorge:
This is a really good catch, thanks for patching these!

PR #196, opened 2024-04-05
Make repo password required

johnmaguire:
Currently, there's no indication that a password is required. If you try to submit the form without one, you get an error flash message:

Screenshot 2024-04-05 at 01-09-41 Backrest

This PR makes it so you can't submit the form until a password is entered (a single character password is accepted.) It also adds a warning if you click off the field without entering a value.

Screenshot 2024-04-05 at 01-07-24 Backrest

Unfortunately I could not figure out how to get the Password label to show the required symbol...

### DIFF
--- a/webui/src/views/AddRepoModal.tsx
+++ b/webui/src/views/AddRepoModal.tsx
@@ -244,6 +244,12 @@ export const AddRepoModal = ({
                   hasFeedback
                   name="password"
                   validateTrigger={["onChange", "onBlur"]}
+                  rules={[
+                    {
+                      required: true,
+                      message: "Please input repo password",
+                    },
+                  ]}
                 >
                   <Input disabled={!!template} />
                 </Form.Item>


StasiumDev:
> This PR makes it so you can't submit the form until a password is entered (a single character password is accepted.) It also adds a warning if you click off the field without entering a value.

Wouldn't that also get rid of the possibility of providing a repo password via the [environment variables](https://github.com/johnmaguire/backrest/blob/repo-pw-required/webui/src/views/AddRepoModal.tsx#L497)?

johnmaguire:
@StasiumDev I had missed that that was desirable. In any case, it looks like the logic you linked is not applying (see the error I received when submitting a form without a password, or env variable.) Perhaps the better fix would be to correct that?

What is the purpose of using the env vars instead of the form field? It is a bit confusing that the form field isn't marked required, IMO.

garethgeorge:
Hey, first off thanks for the batch of pull requests! I'll aim to review these after work / over the weekend.

Re: why password works this way -- this is largely to allow users to provide alternative authentication via env vars like `RESTIC_PASSWORD_FILE` or `RESTIC_PASSWORD_COMMAND`

It might be possible to check the values of the env vars to decide if the password field should be marked as required or not -- or maybe the simplest and most obvious thing is that I can improve the tooltip on the password field to indicate that alternate auth is available.

johnmaguire:
One other option might be to add a checkbox or something like "Use environment variables" to un-require this field. That way, it'd be required by default, but advanced users could make an alternative decision.

Not sure what's best here! I ran into this just when testing, and trying to add a data source quickly. It may not be a real issue in practice.

garethgeorge:
Hmm, I was playing with this issue today. I think there's a related issue going on here, there is validation in place that is supposed to block form submission if env vars for the password aren't set e.g. 

image

but it looks like it's not preventing the submission in this case, I'll followup with a patch to figure out why the validation isn't blocking.

Thinkscape:
An alternative solution would be to ADD form fields, which are mutually exclusive:

- password
- or password file path
- or password command

The validation should be:
- if password provided, valid
- if no password provided, but path not-empty, validate path is a readable file
- if no password, no path, command not empty - try runing password command to retrieve password
- else: invalid

garethgeorge:
Hey, I'll go ahead and close out this pull request as it's out of date :) 

I've also made a number of improvements to how backrest checks that the repo changes actually work and can be connected to before allowing form submission which I *think* addresses the spirit of the issue you were looking at here.

Since this was PR'd backrest now 

 * Verifies the structure of the repo using `restic cat` to verify that the repo's config file exists when submitting new changes to a repo and before actually updating the config in storage.
 * Runs `init` only if the repo doesn't already exist
 * Better plumbing for error messages from those commands (including multi-line error strings) so it's easy to tell why you're not able to connect. E.g. a missing password should be pretty clear!

PR #200, opened 2024-04-06
fix: use C:\Program Files\backrest on both x64 and 32-bit

johnmaguire:
While working on #201, I noticed that this code returned `C:\Program Files (x86)` on 64-bit systems, and `C:\Program Files` on 32-bit systems. However, the code downloads a 64-bit binary on 64-bit machines, so it should go in `C:\Program Files` also.

This PR doesn't try to clean up the old file, in case it is in use by another script on the system.

I do wonder if `C:\Program Files\restic` is the best place to install it - were restic to create an installer, it's plausible they would use that as their own install directory. It may be a little safer for backrest to install it into its own program directory, or perhaps some kind of %AppData% directory. 🤷

### DIFF
--- a/internal/resticinstaller/resticinstaller.go
+++ b/internal/resticinstaller/resticinstaller.go
@@ -242,10 +242,7 @@ func FindOrInstallResticBinary() (string, error) {
 	// Check for restic installation in data directory.
 	resticInstallPath := path.Join(config.DataDir(), resticBinName)
 	if runtime.GOOS == "windows" {
-		programFiles := os.Getenv("programfiles(x86)")
-		if programFiles == "" {
-			programFiles = os.Getenv("programfiles")
-		}
+		programFiles := os.Getenv("programfiles")
 		resticInstallPath = path.Join(programFiles, "restic", resticBinName)
 	}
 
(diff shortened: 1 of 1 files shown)


garethgeorge:
Hey, this is an interesting case. I'll be first to admit that I wish I understood window's permissions model a bit better. iirc I ran afoul of some automatic executable blocking (maybe windows defender?) when executing binaries from `%appdata%`, I should try this again. 

I'm not opposed to putting restic in the backrest program directory though e.g. in `C:\Program Files\backrest` and I think leaving behind the old executable sounds reasonable (could be dangerous to try to clean it up as you point out -- and it's not particularly large).

internal/resticinstaller/resticinstaller.go
```diff
@@ -242,10 +242,7 @@ func FindOrInstallResticBinary() (string, error) {
 	// Check for restic installation in data directory.
 	resticInstallPath := path.Join(config.DataDir(), resticBinName)
 	if runtime.GOOS == "windows" {
-		programFiles := os.Getenv("programfiles(x86)")
-		if programFiles == "" {
-			programFiles = os.Getenv("programfiles")
-		}
+		programFiles := os.Getenv("programfiles")
 		resticInstallPath = path.Join(programFiles, "restic", resticBinName)
```
garethgeorge, line 246: let's also update this to `backrest` in your PR, I like the suggestion to use the backrest directory for the restic dependency as well.

Issue #92, opened 2024-02-14
Restic CLI is faster than Backrest


dhilgarth:
While investigating #86, I copied the exact command that Backrest runs and ran it myself from the command line.
The summary line prints that 197 seconds elapsed. That's 3:17.  
However, Backrest takes 6:25.

Is Backrest reading Restic's STDOUT in a blocking manner and by that slowing down the whole process?

garethgeorge:
interesting report, it may be worth doing this benchmarking in an empty repository. It's very possible that the pack files already exist on the second run if you're rerunning with the same parent and set of files as restic does content-based deduplication. On the second run it's likely nothing is being uploaded.

Backrest uses [this function](https://github.com/garethgeorge/backrest/blob/main/pkg/restic/outputs.go#L92-L150) to read backup progress events in its own go routine. I wouldn't expect it ever be the bottleneck in a backup (though restic does output _a lot_ of bytes of progress events).

dhilgarth:
Unfortunately, my understand of go is only rudimentary. I did however converse with ChatGPT about that and it mentioned that there are other, non-blocking ways of reading from STDOUT: https://chat.openai.com/share/f4d8b443-203f-406b-8130-3c3116eb79f7

My test is with a folder that contains close to 4 million files. No files are changed and it is absolutely repeatable: Running the CLI command takes about half the time of running it in Backrest. It's repeatable, no matter the order.

garethgeorge:
Mind sharing some generic test commands? I'll spend some time benchmarking tonight on an artificial dataset.

dhilgarth:
This is the command I run on the command line:
```
C:\Program Files (x86)\restic> .\restic-0.16.4.exe backup --json --exclude-caches -o sftp.args=-oBatchMode=yes Z:\Projects --tag plan:projects --parent a9839cf1b29e97e3a362cfe6556747d312799655fb19d368d5ae619cc63d5233
```

I've also set the env variables `RESTIC_REPOSITORY`, `RESTIC_PASSWORD`, `AZURE_ACCOUNT_NAME` and `AZURE_ACCOUNT_KEY`.

The final output from the CLI is this:

```
[...]

Issue #180, opened 2024-03-27
Failed to get snapshots issue after leaving Sleep mode on Windows


RobProductions:
**Describe the bug**
I have a scheduled plan that is supposed to run at 5pm each day with the target repo existing on an external NAS machine plugged into my computer. Everything seems to work fine for the most part when the computer is active but I've noticed some backups failed to go off if my computer is asleep during the scheduled time. Of course I wouldn't expect it to run in this state but I saw that it actually attempts a backup after I log in and fails, giving the error `failed to get snapshots for plan` with exit status 1. Because of this I haven't been getting my files saved every time this happens and I have to backup manually using the "Backup Now" button while my computer is awake to get new snapshots.

**To Reproduce**
1. Schedule a plan for a specific time on Windows 11 which backs up to an external drive or NAS (might have to be a NAS machine)
2. Activate sleep mode for the computer before that time
3. Leave sleep mode after the time has passed
4. Failed to get snapshots error shows up and nothing was backed up

**Expected behavior**
Hoping that the backup can succeed once I leave sleep mode and log in.

While I don't know what "failed to get snapshots" exactly means in restic/backrest terms (perhaps it can't connect to the repo location?), my ideas as to why this is happening are:
1. NAS isn't yet mounted properly by the time backrest runs its check after exiting sleep mode OR
2. Some weird security rule is kicking in that won't let programs access the NAS until after I log in on Windows, and backrest is able to run before that point

Some potential approaches for fixing this:
1. Add a (user configured?) arbitrary time delay for X amount of minutes before checking past backups that were missed from Sleep mode in order to allow external devices to mount in time
2. If it's related to logging in, maybe backrest could somehow figure out if the user is logged in or not and hook into that as the best time to run the catch-up operations
3. Have backrest check if the repo is accessible before attempting backup and redo this check every minute until it is able to backup properly, this could be also be user configured for better control. This would ensure that it doesn't just fail and give up, and would also help if someone unplugged a hard drive for example as the next time they plug it in, the backup could still run

**Screenshots**
![image](https://github.com/garethgeorge/backrest/assets/11141862/36fd835d-acc1-45e7-a9e3-1640164f5afe)

Would appreciate any info regarding what exactly this error means as it would help track down the best route to fix it.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Firefox
 - Backrest Version: 0.14.0

[...]

PR #206, opened 2024-04-07
chore: enable full test coverage for Windows

johnmaguire:
After an unsuccessful night debugging #92, I decided to create a benchmark to try to narrow down the cause of the slowness. Unfortunately, I ran into unexplained "Access denied" errors.

I ran the test suite on Windows and saw similar errors in many tests I read on Stack Overflow that AV can be an issue here, so I turned off Windows Defender RTP, to no avail. I think this may be related to the AppData issues mentioned here: https://github.com/garethgeorge/backrest/pull/200#issuecomment-2041206584

To see if this was an issue with my environment, I enabled tests on Github where I discovered the "Access denied" errors do not occur. I have resolved most of the remaining test failures in individual commits for your review.

I'm not sure if there was an actual bug with the `ls` command on Windows, but I received the following error during TestLs`:

```
restic_test.go:261: failed to list directory: command "C:\\Program Files (x86)/restic/restic-0.16.4.exe ls --json 0bca3b482c620f384eae2ca000f7d5ce212171049e5563dd16bc4591345407d7 C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\TestLs1936002979\\002 --no-cache" failed: exit status 1
    Process STDOUT:
    Fatal: All path filters must be absolute, starting with a forward slash '/'
```

I resolved it with a `toPathFilter` command for Windows.

---

However, there seems to be an issue with queue ordering. I am seeing occasional failures in the `TestTaskQueueOrdering` test, and persistent failures in `TestFuzzTaskQueue`. These functions looked a bit dense, so I figured I'd post the PR as-is and see if you had any thoughts on the queue.

---

This PR also enables running the full test suite in Github, to encourage keeping them green. Happy to revert that part if you simply want to improve the existing tests, and continue the current testing strategy.


pkg/restic/path_win.go
```diff
@@ -0,0 +1,37 @@
...
+	"strings"
+)
+
+// toPathFilter converts a directory path to a path filter. Restic requires
+// these to be absolute, starting with a forward slash. On Windows, we convert
+// the drive letter to a single character and prepend a forward slash.
+func toPathFilter(path string) string {
+	cleanedPath := filepath.Clean(path)
+	// If clean results in an empty string, ignoring the volume, it returns "."
+	if cleanedPath != path+"." {
```
garethgeorge, line 17: Hey, not sure I follow what this check is doing -- in general we'll never call toPathFilter with an input that that contains `/../` e.g. `/foo/../`. If we do run into that type of case, I think it'd be fine for that behavior to be unbefined.

pkg/restic/path_win.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	before, after, found := strings.Cut(path, string(filepath.Separator))
+	if !found {
+		// e.g. if path is "C:"
+		before = path
+		after = ""
+	}
+
+	if len(before) == 2 && before[1] == ':' && before[0] >= 'A' && before[0] <= 'Z' {
```
garethgeorge, line 28: If I'm understanding correctly, the goal is to take a path that looks like `Z:/foo...` and convert it to something like `/Z/foo`?

I think we can simplify this a bit e.g.

```
func toRepoPath(path string) string {
   path := filepath.Clean(path) // not quite sure if we need this, we'll always call with absolute paths
   sepIdx := strings.Index(filepath.Separator)
   if sepIdx == -1 || sepIdx != 2 || path[1] != ':' {
      return path
   }
   return "/" + path[0] + "/" + path[3:]
}
```

also renamed to `toRepoPath` which maybe better represents what we're doing e.g. mapping a windows path format to a path as it's stored in the restic repo?

pkg/restic/restic.go
```diff
@@ -306,6 +306,7 @@ func (r *Repo) ListDirectory(ctx context.Context, snapshot string, path string,
 		// an empty path can trigger very expensive operations (e.g. iterates all files in the snapshot)
 		return nil, nil, errors.New("path must not be empty")
 	}
+	path = toPathFilter(path)
```
garethgeorge, line 309: Hey, can you elaborate a bit on what motivated this change? Is this for test coverage or is there also an issue affecting browsing snapshots on windows?

garethgeorge:
Hey, this is awesome! Thanks for tracking the windows breakages down and patching these, left a few questions about the changes in the restic pkg -- mostly wondering what you ran into / what motivated them (and saw some places where we can maybe simplify).

The task queue failures are interesting -- I remember seeing those and better coverage of (and an improved implementation of) the task queue is actually what's currently on my plate as I suspect there are some bugs in there related to https://github.com/garethgeorge/backrest/issues/180. It's possible a related issue is surfacing in the windows fuzztest. 

For now I'm happy to land this test enablement though, do you mind adding t.Skip() to those remaining failures for the time being? :)

Overall this is hugely appreciated -- thanks for contributing it!

pkg/restic/restic.go
johnmaguire: Thanks for checking that on that! I was skeptical this was a real bug, since I hadn't seen errors browsing on Windows. Will update in a bit.

pkg/restic/path_win.go
johnmaguire: It's a weird one... this was for the `C:` test case. [`filepath.Clean`](https://pkg.go.dev/path/filepath#Clean) states that if the path is an empty string, it returns dot. Well... it doesn't include a drive volume in its test. So `C:` will return `C:.` It's odd, tbh: https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/path/filepath/path.go;l=101-103

We can probably ignore this edge case when moving this into the test.

johnmaguire:
Should be good to go! I do wonder if the task queue tests might actually be a time resolution issue? I think in general, Linux has more precise time resolutions than on Windows. I have to run now, or I'd try changing the `Millisecond` reference to `Second` in them.

I skipped the two I've seen fail, but it's possible there are other flaky tests as well. Not sure if it'd help here, but in some of my projects, I enable `-race` detector in CI tests.

garethgeorge:
Time resolution could definitely be at play I've gone ahead and reworked the task queues as a more modular implementation which is easier to cover with tests (and understand), the new coverage seems to pass on windows which is an improvement!

Enabling `-race` detector is a good idea, doing some local runs I'm seeing there are a couple read / write races in the IO handling when executing restic commands, I'll work through those and get some fixes out.

garethgeorge:
Went ahead and submitted a patch for some minor races found by enabling `--race` , this was a great idea. Thanks.

garethgeorge:
Hi John, I'm evaluating relicensing Backrest as AGPL-3.0 . Thanks again for all of the contributions you made early on in Backrest's launch -- I think your changes are some of the most lasting in the code base after a number of refactors :)

I'm wondering if you'd be willing to retroactively [sign a CLA](https://contributoragreements.org/query2form/?_replyto=garethgeorge97@gmail.com&_subject=Contributor%20License%20Agreement%20E-Signing&_body=Fill%20out%20the%20following%20form,%20then%20sign%20your%20initials%20to%20complete%20the%20Contributor%20License%20Agreement.&agreement-type[]=individual&agreement-type[]=entity&fullname=&title=&company=&email-address=&physical-address=&your-initials=&original-agreement=https://contributoragreements.org/u2s/2llymj8bjs&signed-agreement_s=https%3A%2F%2Fcontributoragreements.org%2Fu2s%2F2llymj8bjs%3Fyour-date%3D%40_time%26your-name%3D%40fullname%26your-title%3D%40title%26your-address%3D%40email-address%26your-patents%3D%40Patent-IDs-and-Country_t%26process-url%3D%40_processurl%26action%3Dsign-%40agreement-type%26%40u2s&_processurl=@processurl&_action[0]=https://contributoragreements.org/query2email/&_action[1]=https://contributoragreements.org/query2update/&_next=View%20More%20Contributor%20License%20Agreement%20Signers.&_success=Thank%20you%20for%20using%20contributoragreements.org.%20The%20agreement%20has%20been%20signed%20and%20sent%20via%20E-Mail%20and%20will%20not%20be%20stored.&_submit=Sign%20Your%20Contributor%20License%20Agreement.&_processurl=https%3A%2F%2Fcontributoragreements.org%2Fu2s%2F2lm52x4gui)?

PR #211, opened 2024-04-12
add arch linux installation.

devome:
add arch linux installation in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ services:
 
 ```shell
 paru -Sy backrest
-sudo systemctl enable --now backreset@$USER.service
+sudo systemctl enable --now backrest@$USER.service
 ```
 
 ### Download release
--- a/README.md
+++ b/README.md
@@ -107,8 +107,13 @@ services:
 
 ### Arch Linux
 
+> Note: [Backrest on AUR](https://aur.archlinux.org/packages/backrest) is not maintained by the backrest official and has made minor adjustments to the recommended services. Please refer to [here](https://aur.archlinux.org/cgit/aur.git/tree/backrest@.service?h=backrest) for details. In [backrest@.service](https://aur.archlinux.org/cgit/aur.git/tree/backrest@.service?h=backrest), use `restic` from the Arch Linux official repository by setting `BACKREST_RESTIC_COMMAND`. And for information on enable/starting/stopping services, please refer to [Systemd#Using_units](https://wiki.archlinux.org/title/Systemd#Using_units).
+
 ```shell
-paru -Sy backrest
+## install backrest from AUR
+paru -Sy backrest  # or: yay -Sy backrest
+
+## enable backrest service for current user
 sudo systemctl enable --now backrest@$USER.service
 ```
 
--- a/README.md
+++ b/README.md
@@ -105,6 +105,15 @@ services:
 
 ## Running on Linux
 
+### Arch Linux
+
+```shell
+paru -Sy backrest
+sudo systemctl enable --now backreset@$USER.service
+```
+
+### Download release
+
 Download a release from the [releases page](https://github.com/garethgeorge/backrest/releases)
 
 #### Using systemd with the install script (Recommended)


README.md
```diff
@@ -105,6 +105,15 @@ services:
 
 ## Running on Linux
 
+### Arch Linux
+
+```shell
+paru -Sy backrest
```
garethgeorge, line 111: Is paru the standard way to install packages on arch? Where is the backrest package coming from -- I don't publish one as a part of the release process / not sure if it's kept up to date! I'm vaguely aware a few have been published by 3rd parties but I'd like to contact the maintainers and make sure they're staying up before recommending them in the README.md .

README.md
```diff
@@ -105,6 +105,15 @@ services:
...
+sudo systemctl enable --now backreset@$USER.service
```
garethgeorge, line 112: nit: `backrest`?

same question as above, where does the service definition come from?

garethgeorge:
Thanks! Broadly documenting Arch installation sounds great, some questions though about where the package used here is coming from. I'm not currently maintaining any package registries (other than the new brew tap)

devome:
I am the maintainer of the backrest package on archLinux aur: https://aur.archlinux.org/packages/backrest. If you are willing to personally maintain this package, I can directly add you as the maintainer, so that you can keep updating synchronously. At the same time, you are also more professional about how to build backrest, and you can modify the build process of the package.

Of course, if you don't have time to maintain the aur package, it's not a problem not to accept this PR. If you accept this PR and indicate that it is maintained by a third party, it's okay.

garethgeorge:
First off, awesome that you've gone ahead and created a repo for backrest :)

goreleaser is the tool this project uses for release automation, we can definitely look at setting up backrest to publish to the repo you've setup as a part of the release automation. https://goreleaser.com/customization/aur/ talks a bit about the process.

Re: the service file, thanks for direct linking it -- a few changes I'd suggest

 - it probably shouldn't be setting BACKREST_RESTIC_COMMAND , backrest aims to install it's own version of restic that it's tested against (though I am aware that packages on aur tend to be much more up to date than say the apt version of restic) there are issues with the stability of the JSON API on some versions of restic.
 - since the latest release I've started setting `BACKREST_PORT=127.0.0.1:9898` in the service file, it might be a good practice to adopt here too

I'm happy to either work with you to setup the goreleaser automation or to take a pull request with the options you'd like, I can send you an SSH public key for the goreleaser workflow if you're interested.

devome:
On the one hand, restic is officially maintained by archlinux: https://archlinux.org/packages/extra/x86_64/restic/, and its stability can be trusted. On the other hand, archlinux generally requires the use of local dependencies rather than self downloaded dependencies. So I believe that backrest should rely on the restic in the official repository. If there is really a problem, users can downgrade restic themselves or use `systemctl edit backrest@$USER. server` to unset `BACKREST_RESTIC_COMMAND`.

I will follow your suggestion and add `BACKREST_PORT=127.0.0.1:9898` to `backrest@.servce`.

I have read this link: https://goreleaser.com/customization/aur/ . The link already indicates that "GoReleaser will enforce a '- bin' suffix if it is not present", which means that goreleaser can only create `backlrest-bin` instead of `backrest`. The difference between the two is that the former directly downloads binary files compiled on GitHub for installation, while the latter downloads source code packages and installs them locally after compilation. I didn't create the backlrest-bin` package, but if needed, I can help create it.

devome:
There should be no need to worry about updating backrest on AUR. Currently, I am using nvchecker to automatically update, which will automatically update after the latest release of backrest is released, with a maximum delay of 1 hour.

garethgeorge:
Broadly with those changes this SGTM (and agreed on your reasoning re: BACKREST_RESTIC_COMMAND, I think that's fine for arch).

Left one more comment re: pacman

It'd also be good to add a bit of text explaining what the package is doing in words and explaining how the user can check backrest's status after installation etc. I'll aim to do a docs push this weekend (aiming to setup a proper wiki website) so I can expand on that if you provide some basic text.

Thanks!

devome:
Okay, I have added more explanatory information, please refer to the recent commit.

Pacman is the official package management tool for Arch Linux, which can only install packages from the official repository. Paru/yay is a third-party package management tool based on pacman, which can install packages from AUR.

If my PR still needs to be modified, you can directly modify it.

garethgeorge:
That looks great, thanks for the changes and sorry it’s taken a while to get this merged!

PR #226, opened 2024-04-17
 fix: Linux ./install.sh script fails when used for updating backrest 

eknoes:
When backrest is already installed, and the script is used for updating (as with #187), the install might fail:
If backrest is already running, the cp in the install script fails with "Text file is busy". If the service is already existing, maybe containing customized configuration, it is overwritten. This PR fixes both.

Check if the systemd service is already running, and stop it before installing. Do not overwrite the existing backrest service file.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2,8 +2,14 @@
 
 cd "$(dirname "$0")" # cd to the directory of this script
 
-install_unix() {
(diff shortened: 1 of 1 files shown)


garethgeorge:

eknoes:
Sure, I just signed the CLA

PR #250, opened 2024-04-26
fix: use int64 for large values in structs for compatibility with 32bit devices

zolfariot:
# Motivation
I am a happy user of backrest on a legacy device (NAS QNAP TS-219P).
Restic and back-rest are working fabulously: I manage to convince non CLI-savy member of my family to switch to restic thanks to that UI.
Also, is blazing fast comparing to the official backup UI solutions proposed by QNAP.
However, to make it work I had to patch the output parsing function to avoid overflows in integer variables.
The minor memory and performance overhead of using 64bit vars on 32bit system are unnoticeable, even in my quite low-specs device (512M ram, single-core 1.2GHz arm v5l cpu).

# Commit message

```
fix: use explicity int64 for 32bit compat
For the int type the compiler defaults to int32 on 32bit systems, causing backrest to fail when unmarshaling large number from json.

Here we explictly changed all the int references to explicit int64.
```

### DIFF
--- a/pkg/restic/outputs.go
+++ b/pkg/restic/outputs.go
@@ -55,26 +55,26 @@ type BackupProgressEntry struct {
 	Item   string `json:"item"`
 
 	// Summary fields
-	FilesNew            int     `json:"files_new"`
-	FilesChanged        int     `json:"files_changed"`
-	FilesUnmodified     int     `json:"files_unmodified"`
-	DirsNew             int     `json:"dirs_new"`
-	DirsChanged         int     `json:"dirs_changed"`
-	DirsUnmodified      int     `json:"dirs_unmodified"`
-	DataBlobs           int     `json:"data_blobs"`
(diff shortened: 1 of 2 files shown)


garethgeorge:
Nice, this makes perfect sense. Thanks for contributing the change.

If you have a bit of time to update restic_test.go (should only be a few test cases affected) to match the type changes this LGTM. Otherwise I can go ahead and merge it and patch that test when I have a bit of bandwidth.

garethgeorge:

I'm wondering if you'd be willing to retroactively [sign a CLA](https://cla-assistant.io/garethgeorge/backrest)? Or grant more specific permission to retain the copyright for your PRs but relicense your changes in the codebase under GPL-3.0 .

zolfariot:
Hi! I totally agree with your vision, and I just signed the CLA.

PR #260, opened 2024-05-02
fix: install.sh was calling systemctl on Darwin

jancona:
In attempting to upgrade to 0.17.2 on MacOS, I noticed errors when running install.sh. It appears that the changes in be09303 broke the Darwin install. I restore the `install_unix()` function, which had be renamed, to contain the common code, and left the Linux-specific code in `install_or_update_unix()`. 

I haven't tested on Linux, but the change is small enough that I think I haven't broken anything.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,11 +5,13 @@ cd "$(dirname "$0")" # cd to the directory of this script
 install_or_update_unix() {
   if systemctl is-active --quiet backrest; then
     sudo systemctl stop backrest
-    echo "Updating backrest in /usr/local/bin"
-  else
-    echo "Installing backrest to /usr/local/bin"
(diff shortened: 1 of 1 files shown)


garethgeorge:
Ah, thanks for catching this! LGTM.

I'll also call out, incase it's convenient to you, there's now a brew tap for backrest https://github.com/garethgeorge/homebrew-backrest-tap which is hopefully a more robust way of keeping it up to date. I plan to support both this and the script install. Thanks for the fix!

jancona:
Thanks. I actually noticed the Homebrew option right about when I finished fixing the install script. I switch to Homebrew right after I finished the PR.

garethgeorge:

PR #297, opened 2024-05-25
docs: cookbook example for proxying backrest with caddy

tylwright:
Hey Folks!

I was planning to build an app like this over the long weekend and was pleasantly surprised to find one already exists! I’d love to contribute in any way possible.

I use Caddy as a reverse proxy to front Backrest. I've included a basic example to demonstrate how to front Backrest with Caddy. This setup can be expanded to allow Caddy to auto-generate SSL certificates using the ACME client and to bind to authentication sources like Google via oAuth. This approach simplifies the dev work on Backrest by eliminating the need to implement authentication against Google and other login methods.  I can certainly write that up next!

Thanks for taking this PR into consideration!

### DIFF
--- a/docs/content/3.cookbooks/2.reverse-proxy-examples.md
+++ b/docs/content/3.cookbooks/2.reverse-proxy-examples.md
@@ -1,0 +1,59 @@
+# Reverse Proxy Examples
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


garethgeorge:
Thanks for engaging, contributions very welcome :) glad Backrest is proving a good solution for you and saving you some work!

Reading through the cookbook, this looks great and is exactly what the cookbooks section is for! Caddy is actually what I use as a frontend for my own installation so Backrest's integrated auth support is really mostly for folks using it on a local network :).

LGTM, merging this now.

Issue #120, opened 2024-02-23
Provide alternate image registry to dockerhub


onedr0p:
**Is your feature request related to a problem? Please describe.**

Hi 👋🏼 

Dockerhub has aggressive pull rate limits, it would be awesome if you could push the image to Github Container Registry (GHCR) too.

Thanks!

xane256:
I think the limit of 100 pulls per 6 hours is ok

onedr0p:
> dockerhub has one of the best free tiers for OSS projects that I’m aware of (and iirc GitHub container registry and most alternatives cost money).

GHCR is completely free for open source projects, not sure where you heard otherwise. It's also nice in that it integrates into your repo here.

Take a look at my project [here](https://github.com/onedr0p/exportarr) and look at the packages on the right side. I don't pay Github anything for them hosting my container images.

garethgeorge:
Ah, alright. My understanding was definitely out of date, I had the impression that github's container registry pricing applied to all projects -- I'm now taking it to be private repos only? 

I can go ahead and setup a mirror on github container registry with the next release.

onedr0p:
> I'm now taking it to be private repos only?

Correct which also includes paying for pipeline minutes, which are also free for OSS projects.

christaikobo:
Docker hub has now implemented a harsh 10/hour pull rate limit, which is a quite valid reason to provide an alternative image registry IMO.

Is it possible that the dev can reconsider this?

PR #298, opened 2024-05-25
#120: Publish to GHCR alongside Docker Hub

tylwright:
Taking a stab at publishing to GHCR alongside Docker Hub as requested in #120.  Should be free since this is a public repo.  Apologies if build breaks as I cannot test this out in my env before opening the PR.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -76,6 +76,7 @@ archives:
 dockers:
   - image_templates:
       - garethgeorge/backrest:{{ .Tag }}-alpine-amd64
+      - ghcr.io/garethgeorge/backrest:{{ .Tag }}-alpine-amd64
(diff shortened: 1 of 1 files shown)


garethgeorge:
Hey, thanks for PRing this! 

From my reading at https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry there's a bit more involved that I may need to do / validate on the github actions workflow side of things to ensure the right credentials are available to publish packages. 

I've been holding off on this so far as I haven't seen a huge difference between the ghcr offering and the docker option (though it is my impression that ghcr is preferred for newer projects) though I'm pretty open to trying to mirror I don't want to create too much confusion by publishing in multiple locations.

tylwright:
That's perfectly fair and understandable!  I'll close this for now.

PR #305, opened 2024-05-27
ci(docs): Added broken link detection to docs deploy workflow

tylwright:
This should add a broken link detection step to the docs deployment workflow.  In the background, it's really using [Muffet](https://github.com/raviqqe/muffet) and doing this:
```
muffet -v --max-connections=3 --timeout=20 --buffer-size=8192 --color=always --exclude="(/_nuxt/|discord.com)" https://garethgeorge.github.io/backrest/
```

**Why add this?**
Doc sites are always important to the end user and the devs.  Broken links help no one and can happen out-of-the-blue.  Checking all links during the build would be beneficial as we could fix broken links before users notice them.

**Notes**
* Later on down the road, we may want to have this check be a separate GH Action that runs nightly.
* In this workflow, we're deploying the pages first and then scanning the public endpoint for broken links.  This was done so that the latest docs still get published even if there is a broken link.  We can re-order this later on, if wanted. It also has the ability to scan the local files instead so it can detect any errors _before_ publishing the pages.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -39,6 +39,12 @@ jobs:
         with:
           path: ./docs/.output/public
 
+      - name: Check for broken links
+        uses: ruzickap/action-my-broken-link-checker@v2
+        with:
+          url: https://garethgeorge.github.io/backrest/
+          cmd_params: '-v --max-connections=3 --timeout=20 --buffer-size=8192 --color=always --exclude="(/_nuxt/|discord.com)"'
+
   deploy:
     needs: build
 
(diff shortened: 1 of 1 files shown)


garethgeorge:
Hey! I think this is a great idea, thanks for contributing it. It looks like the workflow runs on the live URL right now which won't actually end up validating the most recent changes (as they're being built). Is it possible to point it at a local filesystem path e.g. the build directory?

tylwright:
I can totally do that!  I'll get that change checked in later this afternoon and append it to this PR.  Thanks!

tylwright:
I've updated it to check the local docs.  Note: The URL arg is required even when providing the local dir's path.

garethgeorge:
LGTM, thanks for setting this up!

PR #319, opened 2024-06-04
fix: rare deadlock in GetOperationEvents

garethgeorge:


internal/api/backresthandler.go
```diff
@@ -221,23 +220,36 @@ func (s *BackrestHandler) GetOperationEvents(ctx context.Context, req *connect.R
 		select {
 		case events <- event:
 		default:
-			errChan <- errors.New("event buffer overflow, closing stream for client retry and catchup")
+			select {
+			case errChan <- errors.New("event buffer overflow, closing stream for client retry and catchup"):
+			default:
+			}
```
Thinkscape, line 226: This is a semi-decent indication of disconnection/unresponsive client - does this case here result in proper socket disconnection?

garethgeorge: Yes -- the logic is a bit convoluted, I'd like to followup and refactor it to use some better golang patterns, but essentially the function is expected to return with the first error sent to errChan (which also disconnects the client), all following values should be dropped on the floor since there's no longer a connected client to send the errors to.

PR #462, opened 2024-09-08
feat: implement rustic support

garethgeorge:
This is largely an experimental exploration into rustic support https://github.com/rustic-rs/rustic


ilkersigirci:
This is huge, when will be see this in the release? Is there any roadmap, thanks

garethgeorge:
Thanks for the status ping on this, it's still something I'm interested in but got lost in the backlog of backrest work :). 

I think my current thinking is to re-evaluate whether I want to create a gRPC daemon that wraps rustic (which is a pretty big maintenance burden unless this is something the rustic project can provide). But I think it should be reasonable for me to wrap rustic's cli (which IIUC already has some json output?) and see how far I can get with that.

PR #474, opened 2024-09-13
fix: windows installation for restic 0.17.1

garethgeorge:


garethgeorge:
Something odd is going on here w.r.t a possible behavior discrepancy between linux and windows in restic as of the patch release? Will need to reach out to the restic devs to understand what the expected behavior is here.

PR #480, opened 2024-09-18
feat: initial support for healthchecks.io

inode64:
- Introduced a new hook to enhance the healthchecks system with more comprehensive lifecycle signals.

https://healthchecks.io/docs/http_api/


internal/hook/types/healthchecks.go
```diff
@@ -0,0 +1,80 @@
...
+		event == v1.Hook_CONDITION_SNAPSHOT_START {
+		PingUrl += "/start"
+	}
+
+	// Send a "fail" signal to healthchecks.io when the hook is failing.
+	if event == v1.Hook_CONDITION_UNKNOWN ||
+		event == v1.Hook_CONDITION_ANY_ERROR ||
+		event == v1.Hook_CONDITION_CHECK_ERROR ||
+		event == v1.Hook_CONDITION_PRUNE_ERROR ||
+		event == v1.Hook_CONDITION_SNAPSHOT_ERROR {
```
garethgeorge, line 46: perhaps my only nit is it may be worth declaring some utilities or constants for this e.g. isStartCondition, isErrorCondition, isSuccessCondition so that we can be consistent about this going forward.

These blocks probably also should have handling for `CONDITION_SNAPSHOT_SUCCESS`, `CONDITION_PRUNE_SUCCESS`, and `CONDITION_CHECK_SUCCESS`

garethgeorge:
Thanks for contributing this! Looks great and I think is very reasonable, my understanding is that one of the most commonly used command hooks is to ping healthchecks.io -- native support is a good idea.

Left a comment re: a more scalable approach to handling the conditions, but overall looks great :)

internal/hook/types/healthchecks.go
inode64: Yes, I already thought about it,
I can create an array with the constants and the function or, a function only using switch like this example

<pre>
func isStartCondition(event v1.Hook_Condition) bool {
	switch event {
	case v1.Hook_CONDITION_SNAPSHOT_START,
		v1.Hook_CONDITION_PRUNE_START,
		v1.Hook_CONDITION_CHECK_START:
		return true
	default:
		return false
	}
}
</pre>

What do you see better?

inode64: I've been looking for a more elegant solution, see what you think.

garethgeorge: Dictionary approach looks good to me!

garethgeorge:
Going ahead and marking as approved, I'm going to hold off on merging until the weekend as this makes sense for Backrest's next feature release (v1.6.0) which will be middle of next month. 

I'm aiming to put out a bug patch release over the weekend that will include the bug fixes in restic 1.17.1 and then start merging the new features for the next minor version release.

garethgeorge (approved):

garethgeorge:
Released in 1.6.0 :)

PR #501, opened 2024-10-02
Exclude apps from jetbrains and extension "Local History" from vscode in .gitignore

inode64:


garethgeorge:
Hey, looks like you've working with backrest locally :) thanks for the interest -- but I'm hesitant to upstream .gitignore changes actually as it's prone to becoming an ever growing list of IDE excludes.

The .gitignore backrest has now is actually fairly out of date and should probably be cleaned up.

PR #518, opened 2024-10-14
feat: add a summary dashboard as the "main view" when backrest opens

garethgeorge:


inode64:
The dashboard is a good idea, and could also be improved by adding:

* the restic version, location, and general status of the system in order to verify that the system is correctly used in healthcheck systems (Monit, Zabbix, etc.)
* Show environment variables, the operating system and the location where the internal database is stored

PR #567, opened 2024-11-18
docs: add Windows PowerShell hook examples

homandr:


garethgeorge:
Thanks for contributing the sample -- this looks great & really neat to see scripting used this way on Windows.

Also motivated me to take another look at why commands are executing inconsistently on windows (needing extra newlines). I think I was missing the `-Command -` flag to tell powershell it's being scripted. Possibly fixed in https://github.com/garethgeorge/backrest/pull/569 but it's quite late so I'll test on my windows install tomorrow.

LGTM and merged.

Issue #549, opened 2024-11-11
Put restic binary in a user-writable location on Windows


homandr:
I would like to suggest that maybe Backrest should store restic binary not in C:\Program Files\Backrest as it does currently, but in a user-writable location. For example, %localappdata%. This is what Chrome browser uses as an example, and Firefox when UAC is cancelled during the installation. C:\Program Files on Windows requires UAC elevation to write even if the current user is an administrator. This creates several issues:

1. Backrest must be launched with elevated privileges on the first run in order to download restic binary. It is mentioned in the documentation but makes Backrest less user-friendly for non-technical users.
2. If restic self-update through Backrest is ever implemented (would be really nice), this UAC requirement makes things more difficult, requiring that Backrest is either executed elevated for each update or always run elevated.
3. A user wanting to use Backrest installer currently must have admin permissions, even if technically it is not needed to use the tool. I know one can just download and unpack zip archive anywhere, but then the user needs to set up shortcuts, basically doing manually what the installer does.

I think even Backrest itself could be stored in %localappdata%. It would eliminate all issues listed above and also make Backrest self-updates (if ever implemented) on Windows easier.

garethgeorge:
Hey -- really interesting and thanks for the fairly detailed input on considering locations (and reasons). This is something I've struggled with a bit on Windows as my understanding of correct packaging for the platform is relatively weak.

I like the idea re: keeping backrest in %localappdata%, assuming the installer should still be able to configure it to launch on startup from there -- and I imagine that should be the case. Doing that, backrest installs restic using a relative path on windows so it would install next to the backrest binary it in the same directory.

Interestingly I also see a `C:\Users\gareth\AppData\Local\Programs` folder on my system where a few programs are storing binaries on my system, I'll learn a bit more about what the conventions are here.

homandr:
Just a quick note - I found out Backrest is actually storing restic binary in the working directory, not necessarily the same directory as Backrest binary. It's just that currently the installer uses C:\Program Files\Backrest in the startup shortcut properties. When I changed that, it re-downloaded restic to the new location.

For multiuser installs you don't want to use %localappdata%. There is %programdata% that's accessible to all users, but not sure if it's appropriate to install binaries there.

Currently the installer creates the shortcuts in the current user's Startup and Desktop folders anyway, so as it stands, it is targeting the current user.

homandr:
[...]

PR #568, opened 2024-11-19
fix: improve windows installer and relocate backrest on Windows to %localappdata%\programs 

homandr:
Various fixes before decisions are made regarding multi-purpose installer (https://github.com/garethgeorge/backrest/issues/549)

1. Changed installation location to user %localappdata%\programs since the installer already installed all icons and made registry entry only for the current user.
2. Fixed blurry installer text on high DPI screens.
3. Fixed icon for Start Menu, Desktop shortcuts, and Add/Remove programs entry.
4. Fixed uninstaller not terminating running application and not fully cleaning up installed files.

Users upgrading from the previous installer should uninstall it first to avoid conflicts. User configuration will stay intact.

### DIFF
--- a/build/windows/install.nsi
+++ b/build/windows/install.nsi
@@ -3,7 +3,7 @@
 !define APP_NAME "Backrest"
 !define COMP_NAME "garethgeorge"
 !define WEB_SITE "https://github.com/garethgeorge/backrest"
-!define VERSION "00.00.00.00"
+!define VERSION "1.6.1.0"
 !define COPYRIGHT "garethgeorge   2024"
 !define DESCRIPTION "Application"
 !define LICENSE_TXT "${BUILD_DIR}\LICENSE"
@@ -29,6 +29,7 @@ VIAddVersionKey "FileVersion"  "${VERSION}"
 
 ######################################################################
 
+RequestExecutionLevel user
 SetCompressor ZLIB
 Name "${APP_NAME}"
 Caption "${APP_NAME}"
@@ -36,7 +37,8 @@ OutFile "${INSTALLER_NAME}"
 BrandingText "${APP_NAME}"
 XPStyle on
 InstallDirRegKey "${REG_ROOT}" "${REG_APP_PATH}" ""
-InstallDir "$PROGRAMFILES64\Backrest"
+InstallDir "$LOCALAPPDATA\Programs\Backrest"
+ManifestDPIAware true
 
 ######################################################################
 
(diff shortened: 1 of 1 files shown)


build/windows/install.nsi
```diff
@@ -99,34 +102,34 @@ WriteUninstaller "$INSTDIR\uninstall.exe"
 !ifdef REG_START_MENU
 !insertmacro MUI_STARTMENU_WRITE_BEGIN Application
 CreateDirectory "$SMPROGRAMS\$SM_Folder"
-CreateShortCut "$SMPROGRAMS\$SM_Folder\${APP_NAME}.lnk" "$INSTDIR\${MAIN_APP_EXE}" "" "${BUILD_DIR}\icon.ico" 0
-CreateShortCut "$DESKTOP\${APP_NAME} Console.lnk" "$INSTDIR\${MAIN_APP_EXE}" "" "${BUILD_DIR}\icon.ico" 0
-CreateShortCut "$DESKTOP\${APP_NAME} UI.lnk" "http://localhost:9898/" "" "${BUILD_DIR}\icon.ico" 0
-CreateShortCut "$SMPROGRAMS\$SM_Folder\Uninstall ${APP_NAME}.lnk" "$INSTDIR\uninstall.exe" "" "${BUILD_DIR}\icon.ico" 0
+CreateShortCut "$SMPROGRAMS\$SM_Folder\${APP_NAME}.lnk" "$INSTDIR\${MAIN_APP_EXE}" "" "$INSTDIR\icon.ico" 0
+CreateShortCut "$DESKTOP\${APP_NAME}.lnk" "$INSTDIR\${MAIN_APP_EXE}" "" "$INSTDIR\icon.ico" 0
+CreateShortCut "$DESKTOP\${APP_NAME} UI.lnk" "http://localhost:9898/" "" "$INSTDIR\icon.ico" 0
```
garethgeorge, line 107: neat!

build/windows/install.nsi
```diff
@@ -3,7 +3,7 @@
 !define APP_NAME "Backrest"
 !define COMP_NAME "garethgeorge"
 !define WEB_SITE "https://github.com/garethgeorge/backrest"
-!define VERSION "00.00.00.00"
+!define VERSION "1.6.1.0"
```
garethgeorge, line 6: Note: this will actually be out of date immediately with the next version, the first version of this installer will ship as 1.6.1. 

I'll merge as is since it's not really a blocker, but I think I need to followup up re: automatically passing in the version from the release workflow.

garethgeorge:
Test failure is a flake -- broadly LGTM. Thanks for contributing this! Per the discussion on the bug I think your recommendation of going with a single user install makes a lot of sense. It probably also has fewest footguns.

Overall quite excited to see backrest's packaging on windows improved, and you're motivating me to look a bit more at a few related windows issues :).

build/windows/install.nsi
homandr: It turns out NSIS is quite flexible and already had such a scenario covered. I will submit a new PR to have the installer grab the version from the changelog.

garethgeorge: Neat! Is that a built in capability of NSIS? My original thinking had been to stamp the version based on the git tag’s on the current commit — the release workflow tags a release commit with the current version.

homandr: Yes, built-in functionality. Given the format and structure of changelog.md, this should work well.

homandr:
Just a few additional comments on the installer. When releasing 1.6.2, you might want to add a note for the users regarding the correct upgrade process. Because the current installer doesn't stop Backrest, it doesn't fully remove it, and also leaves behind Program Files\restic, install.lock, and desktop shortcuts. The cleanest way to upgrade would be:
1. Exit Backrest.
2. Uninstall 1.6.1 (shows up as 00.00.00.00).
3. Install 1.6.2.

It still won't clean up the above items, but at least there won't be a situation with a second Backrest instance launching upon new version installation. Even if a user doesn't follow the process above, the new installer will update the startup shortcut, so after a reboot it will be good. Except for the remnants from the old installer, which won't interfere at that point.

I do need to add process "check and stop" function to avoid multiple processes issue with future upgrades. I was going to look deeper into the upgrade process going forward. It will probably be a week before I have time to work on it.

PR #578, opened 2024-11-25
fix: substantially improve windows installer experience

homandr:
* Refactored the code, removed unused parts, added comments.
* In-place upgrade support:
  - Added a check for the old installer in order to ensure a smooth upgrade.
  - Improved the command for stopping Backrest process.
  - Installer now stops Backrest process before the upgrade.
* Switched to LZMA compression, reducing installer size by ~25%.
* Made desktop and startup shortcuts optional.
* Added information about UI port to help prevent conflicts in multi-user concurrent scenario.
* Installer detects BACKREST_PORT env variable and will create/update shortcuts with the custom port.

Known issue:
The old uninstaller doesn't remove everything in C:\Program Files\Backrest. The new installer cleans up unneeded shortcuts and registry entries but cannot remove anything in Program Files since it runs non-elevated. However, the binaries left there don't interfere with the new installer.

### DIFF
--- a/build/windows/install.nsi
+++ b/build/windows/install.nsi
@@ -10,20 +10,20 @@
 !define MAIN_APP_EXE "backrest-windows-tray.exe"
 !define INSTALL_TYPE "SetShellVarContext current"
 !define REG_ROOT "HKCU"
-!define REG_APP_PATH "Software\Microsoft\Windows\CurrentVersion\App Paths\${MAIN_APP_EXE}"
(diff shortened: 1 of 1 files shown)


garethgeorge:
This looks great -- thanks for continuing to contribute improvements on the Windows side of things here! Also saw your bug report re: the tray and port handling. Will aim to get a patch for that tonight when merging this.

In particular -- correctly stopping backrest on windows update is a _huge_ improvement -- quite excited about that alongside the rest of the quality of life in this PR :)

In terms of release plans, I'm going to try to put out a new version tonight and will include this.

PR #603, opened 2024-12-06
feat: cont'd windows installer refinements

homandr:
* Installer is now versioned - both file name and UI.
* Downgrade is now possible without having to uninstall and install from scratch.
* Clearly indicate whether it is an upgrade, reinstall, or downgrade.
* Remember components page selection from previous installation and honour user deselection of previously selected items.
* Disable installation folder selection for upgrade/reinstall/downgrade scenarios.
* Instead of scaring users with the port information, check listening ports and present the relevant information in case of a conflict.
* Refuse installation with conflicting port, which would result in a non-functioning application anyway.
* Removed PowerShell dependency by switching to native commands.
* Fixed the process termination function where it would not terminate the main backrest.exe process forcefully (just the backrest-windows-tray.exe) after all graceful attempts failed.
* Disable "Open Backrest UI" checkbox if "Start Backrest" is unchecked on the finish page as it wouldn't make sense otherwise.
* Start Menu shortcuts are now part of core installation, not deselectable (simplifies installation experience).

### DIFF
--- a/build/windows/install.nsi
+++ b/build/windows/install.nsi
@@ -6,30 +6,32 @@
 !define COPYRIGHT "garethgeorge   2024"
 !define DESCRIPTION "${APP_NAME} installer"
 !define LICENSE_TXT "${BUILD_DIR}\LICENSE"
-!define INSTALLER_NAME "${OUT_DIR}\Backrest-setup.exe"
 !define MAIN_APP_EXE "backrest-windows-tray.exe"
 !define INSTALL_TYPE "SetShellVarContext current"
 !define REG_ROOT "HKCU"
 !define REG_UNINSTALL_PATH "Software\Microsoft\Windows\CurrentVersion\Uninstall\${APP_NAME}"
-!define REG_START_MENU "Start Menu Folder"
 # Extract version from the changelog.
-!searchparse /file "${BUILD_DIR}\CHANGELOG.md" `## [` VERSION_LOG `]`
-# NSIS requires X.X.X.X format in VIProductVersion. Use it everywhere for consistency.
-!define VERSION "${VERSION_LOG}.0"
+!searchparse /file "${BUILD_DIR}\CHANGELOG.md" `## [` VERSION `]`
+!define INSTALLER_NAME "${OUT_DIR}\Backrest-${VERSION}-setup.exe"
 # User variables.
-Var SM_folder
 Var UIPort
+Var WelcomeTitle
+Var WelcomeText
 Var WelcomePortNote
-Var Prev_version
+Var OldVersion
+Var Cmd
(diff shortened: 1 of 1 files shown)


garethgeorge:
Thanks for taking more time to refine this, the first round of improvements is already a really nice upgrade for the windows UX. I also think that with the version properly embedded we should be able to package backrest for winget now -- I'll go ahead and update the related bug.

Improving the documentation and flow around custom ports was on my todo list, you've beaten me to it and it's a nice touch. I agree that the "always present" warning was potentially alarming for first time users who aren't interested in using backrest with multiple accounts. 

The improved "new install" vs "upgrade" vs "uninstall" flows also look to be a really nice touch.

garethgeorge:
Looking at the build snapshot release, I think I'm seeing a failure due to embedding the version name in the installer binary. The released artifacts are hard coded in a few places. 

https://github.com/garethgeorge/backrest/blob/main/.github/workflows/release.yml#L84-L90

and 

https://github.com/garethgeorge/backrest/blob/main/scripts/generate-installers.sh#L1-L20

homandr:
I think it would be very beneficial to include release version in the filenames for all assets, not just Windows installers. I'm not familiar with Github workflows but see that many other projects have the version. Not sure how they do it.

Regarding Windows installer, probably the easiest would be to pass the architecture to NSIS compiler instead doing the copy/rename. I just added this option. So, you can use this in https://github.com/garethgeorge/backrest/blob/main/scripts/generate-installers.sh#L16:
```bash
    docker run --rm -v $(pwd):/build binfalse/nsis -DARCH="$arch" install.nsi
else
    docker run --rm -e TARGET_ARCH=arm64 -v $(pwd):/build binfalse/nsis -DARCH="$arch" install.nsi
```

Couldn't you then use this for https://github.com/garethgeorge/backrest/blob/main/.github/workflows/release.yml#L88-L90?
```
        with:
          files: |
            ./dist-installers/Backrest*-setup-x86_64.exe
            ./dist-installers/Backrest*-setup-arm64.exe
```
Or possibly just Backrest*.exe.

If you don't want to have the version in the filename, you can just take out ${VERSION} from the OutFile directives. It has no impact on anything other than the filename.

garethgeorge:
I think we should probably actually avoid the version string from the binary name to align it with the other releases and ensure release automation finds it in the expected publishing path.  It will also help my future project to create an auto-updater to have a consistent artifact name that I can fetch from the github releases page :) .

homandr:
Understood. I added another commit to take out the version. Passing the arch is still possible.

garethgeorge:
sgtm! Passing the arch is an improvement. Rerunning the workflow, thanks!

PR #604, opened 2024-12-06
Clarify notes about the port for Windows installation

homandr:
Apologies if I caused confusion around the port behaviour. Running the installer as you described with PowerShell would have the installer pick up the session variable and use the port for UI shortcuts, but Backrest itself would still listen on the default port upon the next reboot.
Also, using : instead of 127.0.0.1: would cause Backrest to listen on all interfaces, thus exposing it to outside access.

The correct way to define a user variable persistently is described in this commit. I also made changes to the installer, including the language around the port. If Backrest supported port definition through a command-line switch (like --port=), then I could add an option to the installer to take user input and create the startup shortcut with the switch. The switch would have to be available in the backrest-windows-tray.exe since this is what starts the software. Not saying it is needed, just explaining what would have to be done to avoid the environment variable steps I described. Unfortunately, Windows doesn't support passing session variables through shortcuts, unless using cmd wrapper script, which is a bit ugly.

Custom port is probably an advanced feature not needed by most users. I think what we have now is very reasonable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,8 +227,8 @@ Download a the Windows installer for your architecture from the [releases page](
 
 The installer will place backrest and a GUI tray application to monitor backrest in `%localappdata%\Programs\Backrest\`. The GUI tray application will start on login by default. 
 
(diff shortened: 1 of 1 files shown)


garethgeorge:
Thanks for catching this!

PR #607, opened 2024-12-09
feat: change payload for healthchecks to text

### LINKED ISSUE #552: healthchecks.io hook should send a plain text message as its status
### Discussed in https://github.com/garethgeorge/backrest/discussions/546

<div type='discussions-op-text'>

<sup>Originally posted by **nesalkiN** November 10, 2024</sup>
The request body for calls to healthchecks looks like this:
Capture_Backrest_-_Immich_–_Healthchecks io_–_Brave_2024-11-10_16-53-48_86046533

This is quite hard to read. The warning via e-mail and notification services connected to healthchecks shows this info when it sends out warnings but you can't easily see in the notifications what's going on when in json.

Can this be configured in the field for template? Please ELI5 if possible. ;)

For other software/solutions it could look like this: (this is from duplicati):
Capture_Duplicati_-_Appdata_OneDrive_–_Healthchecks io_–_B_2024-11-10_16-49-07_62998342

(pics from healthchecks)</div>

herobrauni:
Fixes #552 in my tests.
image

I can barely write go code and have done only minimal checks, so viewers discretion is advised.

In any case thank you a lot for the great tool

### DIFF
--- a/internal/hook/types/healthchecks.go
+++ b/internal/hook/types/healthchecks.go
@@ -1,75 +1,64 @@
 package types
 
 import (
-	"bytes"
-	"context"
-	"encoding/json"
-	"fmt"
-	"reflect"
+    "bytes"
+    "context"
+    "fmt"
+    "reflect"
 
-	v1 "github.com/garethgeorge/backrest/gen/go/v1"
-	"github.com/garethgeorge/backrest/internal/hook/hookutil"
-	"github.com/garethgeorge/backrest/internal/orchestrator/tasks"
-	"github.com/garethgeorge/backrest/internal/protoutil"
-	"go.uber.org/zap"
+    v1 "github.com/garethgeorge/backrest/gen/go/v1"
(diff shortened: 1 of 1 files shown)


garethgeorge:
First, thanks for contributing this fix! Generally looks great.

Question about healthchecks.io -- is there a feature of healthchecks that depends on the JSON text field / @inode64 is there a reason you originally used json payloads with a text field set?

Will this switch break an existing use case?

inode64:
When copying another hook I used the same structures that used json, then I noticed the problem that they were not displayed correctly, I tried using a json to text converter internally in healthchecks, but in the end I think it is better that backrest send the body correctly

garethgeorge:
Sgtm, thanks for the context! Given that, I think I feel reasonably comfortable making this change without a migration / providing a toggle. We can just switch over. 

I’ll mark this as a “feat” as it’s slightly breaking— deserves to be tracked as a minor version change.

PR #625, opened 2025-01-02
Add metric when task fails

titilambert:
Here is an example of `/metrics`
image


garethgeorge:
Must have been a typo that I missed this first time around, good catch and thanks for contributing the fix!

Issue #528, opened 2024-10-22
Restore path on Windows wrong


auanasgheps:
**Describe the bug**
Restore path on Windows is wrong, it kinda duplicates it.

Example:

I need to restore C:\Users\John\Desktop\Photos

I select the folder, then I choose the such path.
The restore happens is such folder, but in the folder "Photos" it has re-created all the structure from the root. There's a C folder, then a Users folder and so on.

**To Reproduce**

1. Pick a folder to be restored
2. Enter the correct path of the folder
3. Observe that inside the destination folder the path is createt from the root.

**Expected behavior**
Restore the files as they are

**Platform Info**

- Windows 10 22H2 x64
- Backrest Version; 1.6.1

garethgeorge:
This does need fixing -- a targeted patch related to this was done for Linux, I need to implement and test something similar for Windows.

garethgeorge:
Relevant lines of code are https://github.com/garethgeorge/backrest/blob/b7abbf9603196e1ed169ebbe81d97d5fa89dea08/internal/orchestrator/repo/repo.go#L309-L320 , which as you can see involve a bit of command crafting. Some work is needed to figure out whether the same or something similar works well on windows.

QuarkZ26:
To add to this, mine just "fails"

The first restore I've tried to do seemingly failed with an "access denied" even though I was restoring some Desktop items into some Desktop subfolder. But then I see it's actually restored to C:\Users\username.

Finally I override the path which was /c/users/etc... to C:\Users\username\Desktop and this time it does work though it's still reported as having failed. The target seems wrong and more for Linux systems

```
[repo-manager] 08:53:24.680Z	debug	restore snapshot	{"repo": "Local", "snapshot": "375487f9644e65c17be477a196dbfbfb67f71855347c9253385f56bdcb2c868f", "target": "C:\\Users\\QuarkZ\\Desktop\\Desktop-backrest-restore-375487f9"}

[restic] command: C:\Program Files\Backrest\restic.exe restore --json 375487f9644e65c17be477a196dbfbfb67f71855347c9253385f56bdcb2c868f -o sftp.args=-oBatchMode=yes --target C:\Users\QuarkZ\Desktop\Desktop-backrest-restore-375487f9 --include /c/users/QuarkZ/Desktop

[restic] {"message_type":"error","error":{"message":"UtimesNano: Access is denied."},"during":"restore","item":"\\c\\users"}
```

garethgeorge:
[...]

PR #631, opened 2025-01-11
fix: restore path on Windows

Nassivera:
Fix restore path on Windows, issue #528  

Now when you restore a path on Windows, the entire directory tree is not recreated but only the desired path.


garethgeorge:
Hey -- thanks for the interest in contributing! Making this consistent with the Linux behavior is a goal, appreciate you taking a look at it.

I'm seeing a windows test fail here: https://github.com/garethgeorge/backrest/blob/0dd360b4973b9f60ba706f869a1a6eb883713afd/internal/orchestrator/repo/repo_test.go#L136C1-L137C1 

A few fixes might be needed for windows -- things that stand out to me is that this code was using path.Dir and path.Base , that might need to change to filepath.Dir and filepath.Base respectively which are the cross platform libraries. 

There might be other issues with how the command is being crafted on windows, unfortunately it's truncated in the error message so it might need some debugging 
image

internal/orchestrator/repo/repo.go
```diff
@@ -311,17 +311,25 @@ func (r *RepoOrchestrator) Restore(ctx context.Context, snapshotId string, snaps
...
 
 	if snapshotPath != "" {
+		normalizedPath := strings.ReplaceAll(snapshotPath, "\\", "/")
+
+		dir := path.Dir(normalizedPath)
+		base := path.Base(normalizedPath)
+		
 		if runtime.GOOS == "windows" {
-			opts = append(opts, restic.WithFlags("--include", snapshotPath))
+		    if dir != "" {
```
garethgeorge, line 320: is this branch now the same as the branch for linux? It may be fine to combine them.

garethgeorge:
Overall looks great, the path normalization makes sense to me. Thanks for digging into this / contributing a fix.

rhamkot:
I'm unsure if I'm doing something wrong on my end, but this looks to still be an issue? If I restore a file to its default path on windows, it restores to C:\DriveLetter\Whatever\Path\To\File.

RiddlemanMVP:
> I'm unsure if I'm doing something wrong on my end, but this looks to still be an issue? If I restore a file to its default path on windows, it restores to C:\DriveLetter\Whatever\Path\To\File.

I'm seeing the same issue...

hbilbo:
I'm still getting the same issue. By default whenever restoring a file which was at C:\path\to\file, backrest creates a nested /C folder under the C drive. It also nests the file under another subdirectory created by backrest. The behavior looks like:

Original: C:\path\to\file.txt
Restore: C:\C\path\to\file-backrest-<id>\file-backrest-<id>.txt

Backrest should restore the file to the original directory, without nested /C and file-backrest-id folders created. @garethgeorge Is this not the intended behavior? It seems like restoring a directory also results in the two extra directories added to the path.

Issue #420, opened 2024-08-18
Add Prometheus `/metrics` support


MrModest:
Thank you for the app! The WebUI looks nice and straightforward! I like it!

**Is your feature request related to a problem? Please describe.**
Even though the app provides some stats in the WebUI itself, it would be nice to be able to fetch metrics and configure custom dashboards in Grafana. 

For example, I'd love to create a dashboard (based on these metrics) that shows a list of snapshots/repositories with info like last timestamp, original backup size, size after deduplication and compression, saved space ratio

garethgeorge:
Hey, I think Prometheus support is definitely something that should be on my roadmap.

Looking into it a bit it looks like the major metric types are: https://prometheus.io/docs/concepts/metric_types/#metric-types

I think that it'd make sense for me to export counters for each repo with names e.g.

 * `repo_<repo ID>_snapshot_count`
 * `repo_<repo ID>_size`
 * etc

And similar for each plan i.e.

 * `plan_<plan ID>_snapshot_count`
 * `plan_<plan ID>_error_count`

etc.

garethgeorge:
Started work on Prometheus metrics in 

https://github.com/garethgeorge/backrest/pull/459/files

Added metrics:

```
	commonDims := []string{"repo_id", "plan_id"}

	registry := &Registry{
		reg: prometheus.NewRegistry(),
		backupBytesProcessed: prometheus.NewSummaryVec(prometheus.SummaryOpts{
			Name: "backrest_backup_bytes_processed",
			Help: "The total number of bytes processed during a backup",
		}, commonDims),
		backupBytesAdded: prometheus.NewSummaryVec(prometheus.SummaryOpts{
			Name: "backrest_backup_bytes_added",
			Help: "The total number of bytes added during a backup",
		}, commonDims),
		backupFileWarnings: prometheus.NewSummaryVec(prometheus.SummaryOpts{
			Name: "backrest_backup_file_warnings",
			Help: "The total number of file warnings during a backup",
		}, commonDims),
		tasksDuration: prometheus.NewSummaryVec(prometheus.SummaryOpts{
			Name: "backrest_tasks_duration_secs",
			Help: "The duration of a task in seconds",
		}, append(slices.Clone(commonDims), "task_type")),
		tasksRun: prometheus.NewCounterVec(prometheus.CounterOpts{
			Name: "backrest_tasks_run_total",
			Help: "The total number of tasks run",
[...]

PR #640, opened 2025-01-17
fix: convert prometheus metrics to use `gauge` type

titilambert:
This PR changes the metrics type to reflect the actual state of the tasks
This is linked to https://github.com/garethgeorge/backrest/issues/420#issuecomment-2586489675
here is an example of the new metrics output
```
# HELP backrest_backup_bytes_added The total number of bytes added during a backup
# TYPE backrest_backup_bytes_added gauge
backrest_backup_bytes_added{plan_id="testplan",repo_id="test"} 2309
# HELP backrest_backup_bytes_processed The total number of bytes processed during a backup
# TYPE backrest_backup_bytes_processed gauge
backrest_backup_bytes_processed{plan_id="testplan",repo_id="test"} 11
# HELP backrest_backup_file_warnings The total number of file warnings during a backup
# TYPE backrest_backup_file_warnings gauge
backrest_backup_file_warnings{plan_id="testplan",repo_id="test"} 0
# HELP backrest_tasks_duration_secs The duration of a task in seconds
# TYPE backrest_tasks_duration_secs gauge
backrest_tasks_duration_secs{plan_id="_unassociated_",repo_id="_unassociated_",task_type="collect_garbage"} 0.002664103
backrest_tasks_duration_secs{plan_id="testplan",repo_id="test",task_type="backup"} 1.632216997
backrest_tasks_duration_secs{plan_id="testplan",repo_id="test",task_type="hook"} 0.002073014
# HELP backrest_tasks_run_total The total number of tasks run
# TYPE backrest_tasks_run_total counter
backrest_tasks_run_total{plan_id="_unassociated_",repo_id="_unassociated_",status="success",task_type="collect_garbage"} 1
backrest_tasks_run_total{plan_id="testplan",repo_id="test",status="success",task_type="backup"} 1
backrest_tasks_run_total{plan_id="testplan",repo_id="test",status="success",task_type="hook"} 1
```

### DIFF
--- a/internal/metric/metric.go
+++ b/internal/metric/metric.go
@@ -18,19 +18,19 @@ func initRegistry() *Registry {
 
 	registry := &Registry{
 		reg: prometheus.NewRegistry(),
-		backupBytesProcessed: prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		backupBytesProcessed: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "backrest_backup_bytes_processed",
 			Help: "The total number of bytes processed during a backup",
 		}, commonDims),
-		backupBytesAdded: prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		backupBytesAdded: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "backrest_backup_bytes_added",
 			Help: "The total number of bytes added during a backup",
 		}, commonDims),
-		backupFileWarnings: prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		backupFileWarnings: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "backrest_backup_file_warnings",
 			Help: "The total number of file warnings during a backup",
 		}, commonDims),
-		tasksDuration: prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		tasksDuration: prometheus.NewGaugeVec(prometheus.GaugeOpts{
 			Name: "backrest_tasks_duration_secs",
 			Help: "The duration of a task in seconds",
 		}, append(slices.Clone(commonDims), "task_type")),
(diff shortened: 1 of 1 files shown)


garethgeorge:
Hey -- thanks for PRing this change. 

One question I have about how this performs is -- gauge metrics give point in time signals of the value of a property as I understand it. But don't signal things like "how many instances of the event have occurred" etc.

Meaning that it'll be possible to construct a dashboard that shows info e.g. for the last run of a task, but I'm not sure how this would work for dashboards tracking stats over time e.g. average bytes added, etc.

titilambert:
> Hey -- thanks for PRing this change. Happy to improve the state of prometheus metrics.
> 
> One question I have about how this performs is -- gauge metrics give point in time signals of the value of a property as I understand it. But don't signal things like "how many instances of the event have occurred" etc.
> 
> Meaning that it'll be possible to construct a dashboard that shows info e.g. for the last run of a task, but I'm not sure how this would work for dashboards tracking stats over time e.g. average bytes added, etc. I'm curious how you see this being used / whether this is easier to consume for people building dashboards on prometheus?

You're right you can not do this can of graph with Prometheus only. In my personal case, I'm using InfluxDB as database. And all the stats that Prometheus can expose, a database (like InfluxDB) can do it easily.
Prometheus is NOT a database. It is recommended to NOT keep long term data (compared to an actual database).

That said, if we use Prometheus only and with stats over time (like today), you're resetting your stats every time backrest restart...

I'm working as devOps for a decade now, and every time I have to use Prometheus, I'm disappointed seeing that it's not a turnkey solution. It always needs more tools to be effective.

So, Maybe we can keep the ones that were there and just adding the gauge ones to have all of them available ?

What do you think ?

garethgeorge:
Hi, sorry for the late reply -- sick the last few days. 

Thanks for the background context, given that info it makes sense to me that it's a bit awkward in either case. One can either use prometheus as a realtime metric w/only gauge values. Or one can awkwardly try to scrape stats over time like we have today. But in neither case is it properly providing a log of events as they happen.

Given that, I'm open to trying the change to gauge metrics and seeing if that generally makes dashboarding easier. I don't consider the exported metrics etc stable yet so I'm open to experimentation. 

If you want to convert this from draft I'm happy to review.

titilambert:
@garethgeorge The PR is already replacing the summary to gauge.
I'm using it for few weeks now, I was able to get an alert when I got an actual error on a backup.

garethgeorge:
Thanks for contributing this, looks great. I'll go ahead and merge these in for the 1.7.2 release.

PR #667, opened 2025-02-08
fix: incorrectly formatted total size on stats panel

ayancey:


garethgeorge:
Thanks for contributing the fix, lgtm!

PR #672, opened 2025-02-18
Added Instructions to Readme to allow connections from 0.0.0.0

W1BTR:
By default, backrest suggests connecting to localhost:9898, however, most Linux systems that I work with are headless, and backrest is being set up from an ssh client.

This pull request adds the following notes to the readme:

If the machine is headless, you will need to run this command (or manually edit `/etc/systemd/system/backrest.service` to enter 0.0.0.0).
```
sudo sed -i 's|Environment="BACKREST_PORT=172\.0\.0\.1:9898"|Environment="BACKREST_PORT=0.0.0.0:9898"|' /etc/systemd/system/backrest.service
```
_Note that this will allow web-ui connections from any device on your network. Always enable user authentication and set a password immediately after enabling this_

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,11 @@ services:
 
 Download a release from the [releases page](https://github.com/garethgeorge/backrest/releases)
 
+For example, for the latest 64 bit Linux release:
+```
+curl -L -O https://github.com/garethgeorge/backrest/releases/latest/download/backrest_Linux_x86_64.tar.gz
+```
+
 #### Using systemd with the install script (Recommended)
 
 Extract the release you downloaded and run the install script:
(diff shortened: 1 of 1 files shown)


W1BTR:
Just modified the instructions to use `systemctl edit`. Also added a command to automatically download the latest release for 64 bit as an example (as before I've had to download the latest release, then upload it to the server Im installing backrest on).

garethgeorge:
Hey -- sorry for the late followup, I went ahead and reorganized the README.md and added this note re: editing the systemd unit while I was at it. Can you take a look and check if it captures what you were hoping / feel free to sync and add anything I've missed?

W1BTR:
No apologies needed, thank you for taking the time to look at it. Those instructions look great to me. Thanks!

Taking the liberty of closing this for now, feel free to re-open if you have more thoughts.

PR #679, opened 2025-02-23
fix: Add Beeep action to protobuf definitions and UI components

johan-scriptdrift:
### Only tested on macOS!

- Add Beeep action to the Hook message in config.proto
- Modify HooksFormList component to include Beeep action
- Update go.mod with beeep and related dependencies

Screenshot 2025-02-23 at 01 12 38

Screenshot 2025-02-23 at 01 16 28


proto/v1/config.proto
```diff
@@ -229,6 +230,14 @@ message Hook {
     string webhook_url = 1 [json_name="webhookUrl"];
     string template = 2 [json_name="template"];
   }
+
+  message Beeep {
```
garethgeorge, line 234: Naming note: it may make more sense to name this after the capability e.g. OsNotification or such rather than the library beeep on the off chance that we need to migrate in the future.

webui/src/components/HooksFormList.tsx
```diff
@@ -347,6 +348,56 @@ const hookTypes: {
...
+              }
+            />
+          </Form.Item>
+          <Form.Item name={[field.name, "actionBeeep", "icon"]}>
+            <Select
+              allowClear
+              style={{ width: "100%" }}
+              placeholder={"Choose icon"}
+              options={[
+                { label: "Information", value: "assets/information.png" },
```
garethgeorge, line 392: Where is this asset defined? Is this something backrest needs to bundle?

proto/v1/config.proto
```diff
@@ -229,6 +230,14 @@ message Hook {
...
+    string template = 1 [json_name="template"];
+    string title_template = 2 [json_name="titleTemplate"];
+    double frequency = 3 [json_name="frequency"];
```
garethgeorge, line 237: Im somewhat ambivalent about accepting a frequency and duration , I wonder if most users mind or would rather have some opinionated default e.g. simply an enum. 

I could see something like

```
enum DeliveryMode {
   SILENT,
   LOW_PITCH_CHIME,
   HIGH_PITCH_CHIME,
}
```
. 

WDYT? are you finding the range of frequency control useful?

garethgeorge:
Really nice contribution, thanks for the interest! And I think quite high value given that most approaches to OS notifications are not cross platform and involve powershell or action script items today. Scripting this is something I see users do frequently.

Using a cross platform library / adding native support makes a lot of sense to me and beeep is a good candidate library. Added a few notes but mostly opinions weakly held :).

proto/v1/config.proto
johan-scriptdrift: Agree, I will update.

webui/src/components/HooksFormList.tsx
johan-scriptdrift: The icon is not available on macOS so I wasn't able to test it easily and thought it would work on Windows and Linux. There is an asset folder in the beeep package and I assumed the icons would be loaded from there. 

I've now tried this on windows and the icons does not seem to work their either. So maybe we should remove this input?

PR #684, opened 2025-02-24
fix: improve exported prometheus metrics for task execution and status

titilambert:
This PR, revert to the previous behavior taskRun, which count the number of tasks run and adds a new metrics to get the last task status.
With the change the name of the metrics will fit more with the actual sense of the values.

### DIFF
--- a/internal/metric/metric.go
+++ b/internal/metric/metric.go
@@ -38,6 +38,10 @@ func initRegistry() *Registry {
 			Name: "backrest_tasks_run_total",
 			Help: "The total number of tasks run",
 		}, append(slices.Clone(commonDims), "task_type", "status")),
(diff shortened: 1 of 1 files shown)


garethgeorge:
Hmm, actually looking a bit deeper at this (already merged so this is fine for now) but I'm seeing that you're essentially using 0 to indicate a OK , 1 to count failures, and -1 to indicate other status? 

This might be a bit fragile -- do you have a screenshot of how you're using this in a dashboard? It's also slightly fragile that I could change how those strings are specified in orchestrator in the future https://github.com/garethgeorge/backrest/blob/8bafe7ea35e990377f96662fc81ccdcc34b4dda6/internal/orchestrator/orchestrator.go#L534 , it'd be easy for this to get out of sync. I'll aim to make that a constant in the future if I remember.

PR #686, opened 2025-02-27
Add gzip to Alpine Docker Container

Sophtli:
I'm using gzip to compress my database dumps and wanted to use the --rsyncable argument. Unfortunately, the included BusyBox version of gzip does not include this argument.
It would be great if the latest standalone version of gzip could be included by default.


garethgeorge:
Hi @Sophtli , sorry but my current thinking is that this is probably out of scope for the official backrest image. Backrest maintains a collection of utilities that are tightly related to backup flows, but I need to draw a cutline to avoid being responsible for both backrest maintenance and maintaining updates for an "ecosystem" of related packages.

Issue #689, opened 2025-03-03
Snapshot listed multiple time when same repo is shared between two servers


ziotibia81:
I have one repository, based on rest-server, and two servers running backrest and using the same repository.
One server show the same snapshot of other server multiple times.

View on server 50-backrest-srv

![Image](https://github.com/user-attachments/assets/e4061355-2eb9-4559-995f-6f680ab20d5e)

View on server 54-backrest-srv

![Image](https://github.com/user-attachments/assets/d3df88e2-1ac1-40f2-80b3-0d3ae1fc1edb)

Reindex has no effects.
Backrest 1.7.2

rhatguy:
I'm seeing a similar thing.  I have a repo that gets backed up to resticker (think Restic command line executable automation).  I then index those snapshots into backrest as a gui.  Every time I index the repo, I'm getting another duplicate listing of the snapshot even though only 1 instance of the snapshot exists in the repo.

Vyerni:
I thought I was going crazy. Glad it's not just me.

garethgeorge:
Thanks for reporting this -- definitely a bug. I'll try to track down what's going on here.

garethgeorge:
I believe the bug is in https://github.com/garethgeorge/backrest/blob/main/internal/orchestrator/tasks/taskindexsnapshots.go#L148 , I recently (accidentally) added logic that filters to snapshots from the current instance when considering what already exists / should be indexed. This is a relatively straightforward fix, but existing duplicates may persist in your history. I'll spend some times thinking about options for deduplication.

rhatguy:
Thanks @garethgeorge .  For my use case I’d be fine to just reinfect the whole repo.  I know that might not be feasible for everyone though.

Vyerni:
Are we saying we might just need to clear local cache, and re-connect the remote repo to clear all the extra "snapshots"? If that's the case, it's not a big deal, again in my case.

Vyerni:
@garethgeorge 

I see this has been address in the 1.7.3 release. 

How best do we proceed with clearing duplicated snapshots in lists?

garethgeorge:
A bit late on this, https://github.com/garethgeorge/backrest/releases/tag/v1.8.0 adds a migration that will delete duplicated entries if you haven't yet reset your local db.

PR #693, opened 2025-03-13
fix: index snapshots incorrectly creates duplicate entries for snapshots from other instances 

garethgeorge:


garethgeorge:
Addresses https://github.com/garethgeorge/backrest/issues/689

PR #714, opened 2025-03-28
Update hook documentation for snapshot start for better details

jrich523:
Add details on how the snapshot start can be used to trigger other backups

I am using this to trigger a DB backup before a snapshot it  taken and wanted to be clear that it completed before the snapshot ran. I manually tested to validate that is the case, but i figured this should be something mentioned in the docs.

I was considering adding some additional docs with an example of how to leverage this.

In my case i ended up mounting the docker sock, and used the command `docker exec -t vaultwarden /vaultwarden backup` to trigger a db backup before the snapshot.

Happy to do another PR with an example like this if you think it makes sense.

Thanks!

### DIFF
--- a/docs/content/2.docs/2.hooks.md
+++ b/docs/content/2.docs/2.hooks.md
@@ -7,7 +7,7 @@ Hooks in Backrest allow you to respond to various operation lifecycle events, en
 Hooks can be triggered by the following events:
 
 ### Snapshot Events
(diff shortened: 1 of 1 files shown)


garethgeorge:
Thanks for the docs contribution, that clarification lgtm! 

And on the examples front, contributions to the cookbooks are always welcome -- an example that shows generically how to mount the docker socket and run a command against a container would be great. I imagine some users are also doing things like looping over docker containers and shutting them down / starting them up again after the backup.

PR #737, opened 2025-04-09
chore: contributor tooling fixes

idpaterson:
This addresses some problems that I had trying to get backrest running with VS Code Dev Containers:
* Some npm dependencies were causing issues, regrettably I don't have logs to share details but since you have a lockfile available I am doing the install with `pnpm` (via `npx`) instead of `npm`
* The backrest.go file had moved
* The dev container was not building the backend since it still used go 1.23
* In a search for all 1.23 I updated the github workflows to go 1.24 as well, can remove those edits if not desireable

It may also be helpful to note somewhere the most recently used versions for go packages since mismatched versions make for excess changes (particularly since the generated files all include the version number of the generator). For my other pull request I rolled back two of the dependencies from latest to avoid unnecessary edits:
```
npm i @bufbuild/protoc-gen-es@2.2.2 -g
go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.35.2
```

There were still a few files generated differently but I just ignored them since they were unrelated to my PR and did not have a version number commented in the generated file.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
       // Update the VARIANT arg to pick a version of Go: 1, 1.16, 1.17
       // Append -bullseye or -buster to pin to an OS version.
       // Use -bullseye variants on local arm64/Apple Silicon.
-      "VARIANT": "1-1.23-bookworm",
+      "VARIANT": "1-1.24-bookworm",
       // Options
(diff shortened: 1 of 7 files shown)


garethgeorge:
Thanks for the interest in contributing, updates look great. Sorry you ran into this :) I should probably be doing a better job of maintaining the dev container configs!

Issue #365, opened 2024-07-01
Retention by count and time period


alaw005:
If backups run every 10 minutes and retention is set by time period eg with hourly = 24 the immediately previous backup from 10 minutes ago is deleted. I want to keep the last 6 (or x) backups and hourly after that, plus any further day, week, month, year setting. 

This could be achieved by allowing rention to be set both by count and by time period at the same time. If retention by count is set to x then that many of the most recent backups would be kept and any additional backups deleted UNLESS those backups would otherwise be kept based on the retention by time period setting. In other words only delete backups that don't meet either retention policy.

garethgeorge:
Interesting, it hadn't ever occurred to me that there's a use case for `--keep-last-n` and `--keep-hourly` etc in combination but this makes plenty of sense and support is relatively straightforward to add though it may take me a few versions to get to this.

Will track popularity of this FR in the meantime for my prioritization.

mitchplze:
+1 for this feature, please. 

Right now I can't use Backrest for anything more frequent than hourly backups, as the last ones just get pruned. :( 

This command is an example of how I would do it on the cli, keeping all backups incl. ones that run every 15 min:

```restic forget --keep-within 24h --keep-daily 90 --prune```

mlusetti:
First of all thanks, backrest is indeed pretty nice!.
This feature is very useful IMHO, keep-within the keep-last are cover very wide and useful combinations.
I would go even further. I would separate the forget command from the backup one.
Why would you run forget right after backup since forget does not reclaim/free space ?

raphpa:
In addition to this I think the retention policy explanation on the web interface is not correct. It says `Snapshots older than the specified time period will be dropped`. That would be the `--keep-within-hourly` `--keep-within-daily` etc. flags of restic.

However backrest is using the  `--keep-hourly` etc. flags which keep the last x snapshots, only counting them _if_ they exist for that time period. 

For example if I specify `Hourly: 48`, expecting that hourly snapshots older than 48 hours would be dropped, but only do 4 snapshots on a day (separated by more than one hour), it will keep all snapshots for the last 12 days instead.

idpaterson:
[...]

PR #738, opened 2025-04-09
fix: --keep-within param to mitigate loss of sub-hourly snapshots 

idpaterson:
Adds a required field when the plan schedule is a cron that hits more than once per hour. This provides a way to keep backups more frequent than hourly and offers a cue to the user that the retention settings are not sufficient for their cron. 

See #365, there were several ideas in the thread for fixing this problem. I settled on this one rather than generic "retention flags" because it allows validation to better prevent user error. There are use cases where `--keep-last-n` would be better as well like on laptops, I'm happy to do a new PR for that approach if it seems better. 

The field is visible and required when the cron hits more than once per hour and defaults to 24h:
> New field

It is removed when the schedule does not hit more than once per hour:
> Field removed when unnecessary

It validates the restic duration pattern:
> Validation error


idpaterson:
Started a poll in the relevant issue but I suspect that `--keep-within` as implemented is not a one size fits all solution. We have discussed the case of intermittent use like a laptop, but even intermittent scheduling can complicate the reasoning around `--keep-within`. If I want to keep the last 2 hours but only want to back up during business hours then when backups start in the morning the last 2 hours of the prior business day are forgotten.

I will take another stab at it this evening using `--keep-last-n` instead.

PR #741, opened 2025-04-11
fix: --keep-last n param to mitigate loss of sub-hourly snapshots

idpaterson:
Adds a field to specify how many recent snapshots to keep before the schedule kicks in to prune them. When the plan schedule is a cron that hits more than once per hour this field is required which provides a cue to the user that this field requires customization for their backup. This provides a way to keep backups more frequent than hourly, can save much older snapshots from schedule-initiated cleanup on intermittent use machines, and offers a cue to the user that the retention settings are not sufficient for their cron. 

See #365, this is the second implementation option. The calculator alleviates my main concern about `--keep-last n` which was how difficult it can be to know how many snapshots are sufficient for my schedule. By calculating the minimum and maximum the user can get an idea of the margin of safety built in to their config.

The field is visible and required when the cron hits more than once per hour. It does not have a default value:
> Screenshot 2025-04-10 at 7 38 39 PM

The calculator helps to con
> Screenshot 2025-04-10 at 8 06 21 PM
vey how big of a timespan is covered:

With very sparse schedules the min and max range seem quite helpful. For a long time I was trying to make an average or median work and it just wasn't producing numbers that made sense in the context of gauging risk. This schedule runs every minute, but only one day a month and any sort of average did not capture the situation like the range:
> Screenshot 2025-04-10 at 8 07 26 PM

Combining weekdays with day numbers is a little dicey but it tries to make a good estimate. These flags are additive in crons, so the following means "midnight on the 1st, 15th, and every Saturday"
> Screenshot 2025-04-10 at 8 08 39 PM

The input is still available but the calculator is removed if scheduling is disabled (i.e. we can't guess when you'll do manual backups):
> Screenshot 2025-04-10 at 7 36 38 PM

It also works for basic day and hour interval schedules.
> Screenshot 2025-04-10 at 7 32 53 PM

The timespan calculation is an inefficient brute force approach that just generates enough (a quantity that varies based on the schedule and keepLastN number) upcoming invocation dates and iterates them to find the min and max. There is probably a much more efficient algorithm if anyone has the brainpower to figure it out.


webui/src/views/AddPlanModal.tsx
```diff
@@ -560,6 +576,100 @@ const RetentionPolicyView = () => {
     }
   };
 
+  const hasBucketedKeepLastN = retention?.policyTimeBucketed?.keepLastN > 1;
+
+  // Calculates the average duration of the last N snapshots
+  const calculateBucketedLastNDuration = () => {
```
garethgeorge, line 582: Hmm, I'm not sure I follow the intent here -- is it to help the user estimate a good value for last n that will be long enough to keep all of their snapshots?

webui/src/views/AddPlanModal.tsx
```diff
@@ -602,67 +712,106 @@ const RetentionPolicyView = () => {
...
+          </Col>
+        </Row>
+        <Form.Item
+          name={["retention", "policyTimeBucketed", "keepLastN"]}
+          label="Latest snapshots to keep regardless of age"
+          validateTrigger={["onChange", "onBlur"]}
+          initialValue={0}
+          required={cronIsSubHourly}
+          rules={[
+            {
```
garethgeorge, line 784: nice, this is a good validation to add-- I suspect this is a significant footgun for many users.

webui/src/views/AddPlanModal.tsx
```diff
@@ -602,67 +712,106 @@ const RetentionPolicyView = () => {
...
+                  throw new Error("Specify a number greater than 1");
+                }
+              },
+              message: "Your schedule runs more than once per hour; choose how many snapshots to keep before handing off to the retention policy.",
+            },
+          ]}
+        >
+          <InputNumber
+            addonAfter={!schedule?.disabled && (
+              <Tooltip
```
garethgeorge, line 796: nit: let's move the tooltip and associated logic into a subcomponent e.g. LastNTooltip , there's a decent amount of nuance here. 

Looks like the intent is a performance optimization-- might make sense to use something like useMemo https://react.dev/reference/react/useMemo

e.g. 

```
useMemo(() => { ... expensive calcuation ... }, [JSON.stringify(retention)]);
```

webui/src/views/AddPlanModal.tsx
garethgeorge: Rereading the PR description I think this makes sense to me actually! It might help if the tooltip text better conveys what it's being computed from e.g.

```{n} snapshots with your schedule {cron} represents an expected retention duration of {lower bound} and {upper bound} days, but may vary with manual backups or if intermittently online```

or some such text.

Comment (garethgeorge) [reply webui/src/views/AddPlanModal.tsx:582]: one more nit: not sure that providing a lower and an upperbound makes sense -- may be more intuitive to simply pick one. Most people are probably concerned with the worst case minimum retention? Upper bound is always potentially unbounded e.g. if it's a laptop that gets unplugged for half-a-year before the next backup

File: webui/package.json [hunk 4]
```diff
@@ -25,10 +25,12 @@
     "@types/react-virtualized": "^9.22.0",
     "antd": "^5.22.2",
     "buffer": "^6.0.3",
+    "cron-schedule": "^5.0.4",
     "cross-env": "^7.0.3",
     "events": "^3.3.0",
     "lodash": "^4.17.21",
     "parcel": "^2.13.2",
+    "pretty-ms": "^9.2.0",
```
garethgeorge, line 33: We can avoid this dependency -- there's a `formatDuration` utility in the webui already that takes a time in milliseconds https://github.com/garethgeorge/backrest/blob/fc4bdc089ad456defc38f52b00ff3c5c2a566fda/webui/src/lib/formatting.ts#L63

webui/src/views/AddPlanModal.tsx
```diff
@@ -560,6 +576,100 @@ const RetentionPolicyView = () => {
...
+      ? days.length
+      : Math.floor(
+        // Estimated number of weekdays enabled per month
+        weekdays.length * 31 / 7 
+        // Plus the number of days enabled per month, reduced by the number of days that are also
+        // enabled as weekdays to avoid double counting
+        + days.length * (7 - weekdays.length) / 7
+      );
+    // Get a sufficient number of dates to calculate durations.
+    const dates = cron.getNextDates(Math.max(
```
garethgeorge, line 616: Okay -- reading this more I'm getting a bit caught up in what this is trying to measure and the logic for estimating the sample count.

I think the case we're dealing with is essentially-- what is the "min duration retained" and "max duration retained" with degenerate cases e.g. we run a backup on 1 day of a month and retention is set to 1.

In that case, we're keeping between 0 days of backups and 30 days of backups?

Great, how do we measure that -- my attempt looks something like 

```
export const computeMinAndMaxRetention = (cronExpr: string, samples: number) => {
  const cron = parseCronExpression(cronExpr);
  const itr = cron.getNextDatesIterator();

  const times: Date[] = Array.from({ length: samples }, () => itr.next().value!);
  const firstDate = times[times.length - 1];

  let minSpan = Infinity;
  let maxSpan = 0;

  let simulateDuration = 24 * 60 * 60 * 1000;
  if (cron.days.length !== 31 || cron.weekdays.length !== 7) {
    simulateDuration *= 31;
  }

  let i = times.length - 1;

  while (true) {
    const cur = itr.next().value!;

    if (i > samples && cur.getTime() - firstDate.getTime() > simulateDuration) {
      break;
    }

    const dropped = times[(i + 1) % times.length];
    times[i % times.length] = cur;
    const span = cur.getTime() - dropped.getTime();

    if (span < minSpan) {
      minSpan = span;
    }
    if (span > maxSpan) {
      maxSpan = span;
    }
    i++;
  }

  return [formatDuration(minSpan), formatDuration(maxSpan)];
}

// try min and max for a cron that runs only on the 1st and 2nd of the month
console.log("0 0 1,2 * *", computeMinAndMaxRetention("0 0 1,2,3 * *", 3));
// try min and max for a cron that runs only on monday
console.log("0 0 * * 1", computeMinAndMaxRetention("0 0 * * 1", 2));
// try min and max for a cron that runs only on the 1st of the month and monday
console.log("0 0 1 * 1", computeMinAndMaxRetention("0 0 1 * 1", 2));

```

which outputs

```
0 0 1,2 * * [ '2d0h0m0s', '1mo1d0h0m0s' ]
0 0 * * 1 [ '7d0h0m0s', '14d0h0m0s' ]
0 0 1 * 1 [ '3d0h0m0s', '14d0h0m0s' ]
```

garethgeorge:
Left a few comments, especially interesting algorithmic component re: estimating and displaying how many snapshots are retained.

Really appreciate the contribution and the thought you're giving to the UX.

I think both this and the `keep-within-duration` approaches have their pros and cons, ideally I can see benefits of both. 

I'm probably leaning towards merging `keep-last-n` as I think it lets you do things that are in some ways "more different" from what one gets out of keep-within-duration. If you want to retain some of what keep-within-duration aims to fix (e.g. running more than 1 backup per hour and losing snapshots) I think a new default behavior of specifying `keep-within-duration 1h` unless keep-last-n is specified might be reasonable. WDYT?

webui/src/views/AddPlanModal.tsx
idpaterson: That is a clever way to avoid generating unnecessary dates. I'll give it some thought this weekend, I suspect that focusing on just the minimum duration should make it simpler to write an efficient algorithm to replace this brute force date sampling.

idpaterson:
I haven't gone back to the UI yet to hook it up but I worked up a more efficient algorithm that does not rely on the brute force next date generation. Instead it starts at the minutes working backwards up to days. At each jump (from minutes to hours then from hours to days) it calculates which value to start at in order to minimize the distance between the number of values we need. That calculation varies depending on whether there are consecutive values in the next level up. If the export is enabled for two hours in a row then the minutes can wrap. For example, "0-9,59 * * * *" would start at 0 if wrapping is disabled but 59 is the best starting point if those values can wrap.

It is no longer tied to actual calendar dates so won't produce "weird" results when crossing daylight savings time boundaries. It should probably run through the actual number of days each month, but for now just repeats every 31 days.

I set up a dozen cron expressions in vitest to make sure this agreed with our other algorithms at least until you get into multiple month spans. I don't plan to add vitest tooling to the backrest repo but if you think it would be valuable and want to set up the framework for that I can add the tests.

This still uses cron-schedule for its parser, but it should be possible to get the same data out of react-js-cron to eliminate the dependency. It looks like [`setValuesFromCronString`](https://github.com/xrutayisire/react-js-cron/blob/3666e663b5f2b5a37d2907d672493f6a601808dd/src/converter.ts#L24C1-L99C2) could work. Is there an idiomatic low overhead React way to get values out of that function signature or do I just have to stub out all of those arguments to hijack the parsed values? Unfortunately the `parseCronString` function that it uses internally is not exported.

This is pretty gnarly to match the output of cron-schedule:
```ts
const parseCron = (cronExpr: string) => {
  const noop = () => {};
  const cron: Record<string, number[]> = {};
  // Assign the parsed values, replacing empty arrays for * with all enabled values
  const collectValues = (prop: string, length: number, start = 0) => (value: SetStateAction<number[] | undefined>) => {
    if (Array.isArray(value)) {
      cron[prop] = value.length 
        ? value
        : Array.from({ length }, (_, i) => start + i);
    }
  };
  converter.setValuesFromCronString(
    cronExpr,
    noop,
    noop,
    'always',
    { current: '' },
    true,
    {} as Locale,
    false,
    collectValues('minutes', 60),
    collectValues('hours', 24),
    collectValues('days', 31, 1),
    noop,
    collectValues('weekdays', 7, 1),
    noop,
  );
  return cron;
}
```

idpaterson:
Quick and dirty update to the UI to allow testing the new algorithm. I need to update `formatDuration` to return days and add options for the minimum unit (to skip seconds) and to skip 0. Currently just removing those with a regex for the POC. Will do a separate component and display this differently.

The new algorithm calculates based on `keepLastN - 1` because if you back up every hour and keep 2 snapshots you don't have 2 hours worth of backups. The `forget` happens after the new snapshot is recorded so the only conclusion that makes sense is that you have 1 hour of backups retained. Your algo got this right. I think this can be confusing, may reword the form to clearly ask how many _old_ snapshots the user wants to retain, then `+ 1` to the user's choice for the `--keep-last` arg.

One possible issue with the new algorithm is that if the start point for minutes is chosen based on being able to wrap to the next hour there is no guarantee that the chosen start hour will have a consecutive date following it to make use of that minute wrapping (same applies to hours and days). This might not be an important enough edge case to justify more work on this.

react-js-cron has been updated for the maintainer's speedy release of my PR to export the cron parsing function to avoid that messy code in my last reply. As a bonus this version also fixes the UI glitch when "every" value is selected for a unit:

Before (5.0.1)
> Screenshot 2025-04-18 at 9 14 49 AM 
After (5.2.0)
> Screenshot 2025-04-18 at 9 14 34 AM

garethgeorge:
Apologies for the late review, heavy travel month on my end. Really want to call out that I appreciate all the thought you're giving to this.

Overall I think the new algorithm impl in a separate file looks great. 

> Quick and dirty update to the UI to allow testing the new algorithm. I need to update `formatDuration` to return days and add options for the minimum unit (to skip seconds) and to skip 0. Currently just removing those with a regex for the POC. Will do a separate component and display this differently, this PR is not finished yet.

Absolutely feel free to update the formatDuration -- if it's not dropping 0 duration segments today that's likely a mistake. I think `0` duration should be displayed as `0s` elsewise any unit displayed should be non-zero. I may find time tonight to get a commit in fixing that for you.

> The new algorithm calculates based on `keepLastN - 1` because if you back up every hour and keep 2 snapshots you don't have 2 hours worth of backups. The `forget` happens after the new snapshot is recorded so the only conclusion that makes sense is that you have 1 hour of backups retained. Your algo got this right, my original implementation was wrong. I think this can be confusing both for the user and for `- 1`'s floating around the code, may reword the form to clearly ask how many _old_ snapshots the user wants to retain, then `+ 1` to the user's choice for the `--keep-last` arg.
> 
> One possible issue with the new algorithm is that if the start point for minutes is chosen based on being able to wrap to the next hour there is no guarantee that the chosen start hour will have a consecutive date following it to make use of that minute wrapping (same applies to hours and days). This might not be an important enough edge case to justify more work on this.

+1, I think the inclusion of the algorithm to help users estimate what will be retained already goes above and beyond the MVP for plumbing in `keep last n` retention. What's included now is great attention to detail, anything more we should definitely start considering maintenance / complexity.

> react-js-cron has been updated for the maintainer's speedy release of my PR to export the cron parsing function to avoid that messy code in my last reply. As a bonus this version also fixes the UI glitch when "every" value is selected for a unit:

Very neat that you got this upstreamed, appreciate that this won't add to the backrest codebase maintenance burden :) 

Overall, LGTM happy to merge in current form -- please let me know when you feel it's ready for merging and I'll do a final pass!

idpaterson:
Thank you for your patience as well, life happens and wrapping this up keeps getting pushed off longer than I would like! Looks like I have some uncommitted code to do `formatDuration` implemented how you describe so I'll push that now and hopefully wrap up the UI this weekend.

idpaterson:
I'm done tinkering, feel free to adapt as needed.

garethgeorge:
Released this change in 1.8.1 🎊

PR #757, opened 2025-04-29
[Docker] Add docker-cli-ompose to the base image of the container

larsb24:
Proposal to add docker-cli-compose to the packages the backrest docker image ships by default.

I think it is useful to have access to the docker compose commands in hooks to stop and start whole stacks from backrest.

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -1,6 +1,6 @@
 FROM alpine:latest
 LABEL org.opencontainers.image.source="https://github.com/garethgeorge/backrest"
-RUN apk --no-cache add tini ca-certificates curl bash rclone openssh tzdata docker-cli
+RUN apk --no-cache add tini ca-certificates curl bash rclone openssh tzdata docker-cli docker-cli-compose
 RUN mkdir -p /tmp
 COPY backrest /backrest
 RUN /backrest --install-deps-only


garethgeorge:
I'm a bit hesitant to merge this due to possibly increasing the image size, would you mind posting the size of the built image before and after including docker-compose ?

xd003:
> I'm a bit hesitant to merge this due to possibly increasing the image size, would you mind posting the size of the built image before and after including docker-compose ?

Is it possible to have an environment variable where users can define the specific packages they need, which would then be installed at runtime? This approach would allow users to install only the packages they require nothing more, nothing less resulting in a more streamlined and efficient setup.

jrich523:
`220MB` -> `283MB` ~ `63MB` diff

probably not a crazy difference, but it does seem a bit crazy its 63MB for that one tool :)

on the note of size (without the compose cli)

`-ldflags="-s -w"` -> `209MB`
`upx --best /app/backrest` -> `191MB` (pretty damn slow/costly)

garethgeorge:
Thanks for doing the size comparison -- yeah the binary size bloat from adding the `docker cli` on its own surprised me a bit.

I'm definitely hesitant to keep the set growing, I think the current set is a pretty decent compromise. For users needing further customization, I'll aim to add a new guide to the cookbooks section.

To use a custom image one wants a compose setup something like:

```
  backrest:
    build:
      context: ./backrest  # Path to the directory containing your Dockerfile
      dockerfile: Dockerfile # Name of your Dockerfile (default is Dockerfile)
    container_name: backrest
    hostname: backrest
    volumes:
      - ./backrest/data:/data
      - ./backrest/config:/config
      - ./backrest/cache:/cache
      - ./backrest/tmp:/tmp
      - /path/to/backup/data:/userdata  # Mount local paths to backup
      - /path/to/local/repos:/repos     # Mount local repos (optional for remote storage)
    environment:
      - BACKREST_DATA=/data
      - BACKREST_CONFIG=/config/config.json
      - XDG_CACHE_HOME=/cache
      - TMPDIR=/tmp
      - TZ=America/Los_Angeles
    ports:
      - "9898:9898"
    restart: unless-stopped
```

and then a docker file that looks like

```
FROM garethgeorge/backrest:latest
RUN apk add docker-cli-compose
```

larsb24:
> I'm a bit hesitant to merge this due to possibly increasing the image size, would you mind posting the size of the built image before and after including docker-compose ?

Hi, yeah I actually noticed that the proposal wasn't a good idea afterall because trying to launch a docker compose from inside another docker container has the issue that the user of the other container (backrest in this case) is the one launching it and you would have to pass through all the volumes of the other container to backrest as well otherwise the service runs without any data.

I solved my issue by just creating an ssh key, copying it to my main user on the server and mapping the .ssh directory as a volume in the compose file. That way I can now start and stop my other compose stacks with ssh.

PR #760, opened 2025-04-30
feat: enhance environment variable handling in NewRepo, allow the rep…

adrian-taralunga:
feat: enhance environment variable handling in NewRepo, allow the repo uri to have envirnonment variables, prioritising first the ones from opts and then adding if any missing from the os environment

### DIFF
--- a/pkg/restic/outputs.go
+++ b/pkg/restic/outputs.go
@@ -125,56 +125,105 @@ func (b *BackupProgressEntry) Validate() error {
 	return nil
 }
 
-// readBackupProgressEntries returns the summary event or an error if the command failed.
-func readBackupProgressEntries(output io.Reader, logger io.Writer, callback func(event *BackupProgressEntry)) (*BackupProgressEntry, error) {
+func (b *BackupProgressEntry) IsError() bool {
+	return b.MessageType == "error"
+}
+
+func (b *BackupProgressEntry) IsSummary() bool {
+	return b.MessageType == "summary"
+}
+
+type RestoreProgressEntry struct {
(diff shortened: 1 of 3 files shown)


garethgeorge:
Thanks for the interest in contributing, generally open to supporting more expansions with some caveats --

I've been hesitant to blanket expand $ENV syntax as is done in shells as this may conflict with $ in a user's password or in an env var value or secret-- I don't want people to have to commonly escape values. I've settled at the moment for substituting anything of the form `${VAR}` as the only supported substitution format.

The existing env var expansion is implemented here https://github.com/garethgeorge/backrest/blob/3ab62710a1e2278ab93479d0204fd86d33d436dd/internal/orchestrator/repo/env.go#L13

But isn't applied in as many places as it could be, env var substitution actually happens a layer above the package you're looking at. It happens here https://github.com/garethgeorge/backrest/blob/3ab62710a1e2278ab93479d0204fd86d33d436dd/internal/orchestrator/repo/repo.go#L36-L83 . The thought being that the restic package should avoid business logic and focus on the contract w/restic. 

So the concrete changes I'd make are 

 * use ExpandEnv util linked
 * apply ExpandEnv to the repo URI on this line https://github.com/garethgeorge/backrest/blob/3ab62710a1e2278ab93479d0204fd86d33d436dd/internal/orchestrator/repo/repo.go#L76C37-L76C47

adrian-taralunga:
@garethgeorge  thank you for you comment. this is my first time using go, so I really appreciate your directions. I created another pull request with the changes you recommended. I see it's working on my side.

PR #799, opened 2025-06-09
Fix ToolTipIcon Info Enum

4fosaefp:
The [ToolTipIcon](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.tooltipicon) information enum is "Info" rather than "Information"


garethgeorge:
Hi @4fosaefp , apologies for the late review. Appreciate the fix to the installer script, merging.

PR #813, opened 2025-06-21
feat: allow repo url to be set from env with ${RESTIC_REPOSITORY}

adrian-taralunga:


garethgeorge:
Thanks for the contribution -- expanding the URI w/env var substitution sgtm. Revisions look great!

adrian-taralunga:
Thank you for creating and putting so much work into backrest.

PR #827, opened 2025-06-30
feat: update windows installer to support run as administrator at login

garethgeorge:


homandr:
@garethgeorge, I noticed you started working on this. I spent quite a bit of time a few months ago on this and could attempt to finish it. Where I left it was pretty much working installer but with a number of caveats that I didn't find a good solution at the time. Mostly around dealing with uninstalling, upgrades, detecting single-user vs multi-user installations. There are a lot of corner cases that perhaps don't need to be addressed if they affect 1% of users. I was just trying to be as thorough as I could.

If you don't mind waiting for a couple of months, I think I could get back to working on this. I see more and more requests along these lines.

garethgeorge:
Hi @homandr really appreciate your continued interest in looking at this -- yep started this PR as a brief attempt on my end to see what'd be involved in adding the ability to install as administrator, this PR is largely LLM generated. Afraid I was struggling to understand whether it's correct without gaining more context in nsis and I believe it attempted to use some features that aren't available. 

All of which is to say, I'd be very appreciative if you're interested in taking this on, I'm impressed with the improvements you made so far to the Windows installer.

garethgeorge:
Hi Homandr, apologies

PR #828, opened 2025-06-30
fix: race condition in taskcollectgarbage potentially prematurely deletes logs for tasks currently running

garethgeorge:


garethgeorge:
note: the only place where backrest allows parallel command execution today is the `Run Command` feature, so this is a very rare situation.

PR #836, opened 2025-07-07
docs: add rclone mount to docker-compose example

poiintt:
Just a small documentation change to the README to give an example on how to persist the rclone configuration when using Docker


garethgeorge:
Looks great, figuring out this path is probably a common pain point for people setting up rclone. Merging.

PR #867, opened 2025-08-11
fix: modernize windows installer with Inno Setup

homandr:
I built a new installer from scratch using Inno Setup. I was struggling with NullSoft to do what I wanted and decided to take a look at Inno. It supports administrative and non-administrative install modes out of the box. It's so much easier to work with, there are countless built-in powerful and convenient features which would be impossible or very difficult in NSIS. Very good defaults. Custom scripting is much clearer and easier. Installer icons have higher resolution. Includes IDE and debugger for easier development.

Since there are a lot of changes with the new installer, I would like to explain the details and some of the decisions.

**Welcome and license**
I disabled the welcome page (actually, it's disabled by default in IS), since there is no value in it. And Microsoft recommends disabling it: https://learn.microsoft.com/en-us/windows/win32/uxguide/win-wizards#wizard-pages

I also disabled the license page. The GPL license text says it isn't required to be accepted (paragraph 9), and since you already don't require acceptance for non-installer downloads, I don't know why the installer should be different. Let me know what you think. It's just one line of code to enable it back.

**ARM64**
Inno Setup can produce a unified installer that runs on both x86_64 and ARM64 and has configurable checks to deploy the appropriate binaries. I tried the unified installer and found out that the installer size is pretty much doubled (makes sense). It also might be more understandable for users to have separate installers. So I kept the original approach of unique installer for each platform. Let me know if you would prefer otherwise.

**Install modes**
The new installer supports three install modes:
1. User-specific non-elevated.
2. System-wide elevated, running under the system user context.
3. System-wide elevated, running under the current user context.

With Backrest being a web-based application and also one that runs in the background, there is an almost infinite number of other variations when it comes to multiple instances. I didn't target all of them. In fact, I had to write extra code to prevent some undesirable scenarios, for example, `#1` and `#3` for the same user, as it would result in one instance being broken anyway (Backrest configuration location is the same in these cases).

The `#1` and `#2` were fairly easy to implement, they are pretty much standard modes. But I made an extra effort to add `#3` as I think it's the best approach for anyone wanting to run elevated. It retains user security context, making it possible to access CIFS shares as an example. The downside is that it's really for a single-user scenario, although Backrest is still obviously accessible to other users as long as the original is logged in.

**Elevated privileges**
I used Windows Task Scheduler to achieve this. We previously discussed using a third-party tool to run Backrest as a service, which would come with its bag of issues. While TS might not be the perfect answer, it's a native Windows tool used by many people in the exact same scenario without issues. On top of that, `#3` is really only achievable through TS (although services can be run under any account, it would require the user to enter their password there).

**Multi-user installations**
The three modes above should cover all needs for a single-user scenario. For multi-user scenario the only fully supported option is `#1` with one caveat: for concurrent use each user must select a different network port during the installation. The installer checks if the selected port is in use and instructs the user to select a different port. The default 9898 doesn't add any arguments in the shortcuts, allowing to use the environment variable (which would otherwise be overridden by the --bind-address).

For non-concurrent use ports don't matter. Both administrative options `#2` and `#3` are system-wide installations and use port 9897 in order not to overlap with user mode. Only one system-wide installation is allowed, either `#2` or `#3`. This is not only because such install mode is "wired" into IS to be system-wide and would require a lot of custom code to change that, but also for security reasons. We don't want to run an executable with elevated privileges with the executable located in non-admin-writeable directory.

If some users install multiple instances with `#1` and then use the Task Scheduler approach to run them elevated, they are free to do so on their own. They can even use the task from `#2` as an example of how it's done. But the installer will not deploy an insecure configuration.

**Old installer**
The new one will detect the old installation and prompt the user to uninstall it. It was cleaner to do it this way instead of trying to "take over" and clean up after the old installer.

**Summary of changes**
1. New installation builder.
2. Administrative install modes.
3. Port selection option. Changing the port for an existing installation is possible by re-running the installer.
4. Uninstaller can optionally delete Backrest configuration.
5. Option to add Backrest installation directory to environment PATH.

For future reference—since I don't often visit Backrest repository page—if there are issues raised about the installer, feel free to mention me in the relevant thread.

P.S. For some reason Github-compiled installer has a slightly different look from what I have on my PC. It's probably to the runner having a slighly older version of Inno Setup.

Github:
image

My PC with 6.4.3:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,13 +55,36 @@ jobs:
             dist/*.tar.gz
             dist/*.zip
 
-      - name: Generate Installers
-        run: |
-          mkdir -p dist-installers
-          ./scripts/generate-installers.sh ./dist-installers
+  installer:
+    name: Windows installers
+    needs: [ build ]
+    runs-on: windows-2022
 
-      - name: Upload Installers
-        uses: actions/upload-artifact@v4
-        with:
-          name: backrest-snapshot-installers
-          path: dist-installers/*.exe
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Download build artifacts
+      uses: actions/download-artifact@v4
+      with:
+        name: backrest-snapshot-builds
+
+    - name: Unzip artifacts and compile Inno Setup installers
+      shell: powershell
+      run: |
+        mkdir windows_installers
+        foreach ($arch in "x86_64", "arm64") {
+          $src = "backrest_Windows_$arch"
+          Expand-Archive ".\${src}.zip"
+          cp build\windows\* $src
+          & "c:\Program Files (x86)\Inno Setup 6\ISCC.exe" /DArch=$arch ${src}\installer.iss
+          cp "$src\Output\*" windows_installers
+        }
+
+    - name: Upload installers
+      uses: actions/upload-artifact@v4
+      with:
+        name: backrest-snapshot-installers
+        path: windows_installers\*.exe


build/windows/install.nsi
garethgeorge, line 1: Switching off of nsi sgtm :) it was starting to leave a bad taste in my mouth as well.

.github/workflows/build.yml
```diff
@@ -55,13 +55,36 @@ jobs:
...
+      with:
+        fetch-depth: 0
+
+    - name: Download build artifacts
+      uses: actions/download-artifact@v4
+      with:
+        name: backrest-snapshot-builds
+
+    - name: Unzip artifacts and compile Inno Setup installers
+      shell: powershell
```
garethgeorge, line 75: nice, I think overall switching to and actually using windows/powershell to build the windows installer is a nice touch and makes sense.

build/windows/installer.iss
```diff
@@ -0,0 +1,352 @@
...
+  end;
+end;
+
+function GetPortParam(Param: String): String;
+var
+  S: String;
+begin
+  S := GetPort('');
+  // Don't add any shortcut parameters for default port selection.
+  if S = '9898' then Result := '' else Result := '--bind-address 127.0.0.1:' + S;
```
garethgeorge, line 160: Looks like I'll need to get fixes to the backrest mon binary to open the WebUI properly at a user configured bind addr.

garethgeorge:
Overall looks great, I left a few commends and will do a deeper dive later in the week to try this out and also get in some of the dependent fixes you need (mostly the bind-address issue, please let me know if there's anything else that stands out). but wanted to get an initial pass out :)

Re a few of your questions / decision points

 1. I agree that dropping the license display from the installer is absolutely fine. 
 2. Agree with your decision re: separate installers, it's more explicit.
 3. Install modes, just want to callout that I think the explicit support you've given for a system level and user level install running side by side on a single host is a really nice capability.  Can you explain a bit about where the various modes will store data / how you prevented overlapping? And particularly interested to learn a bit about the problems mode 3 "system level with current user context" solves.
 4. It's a bit late and I need to dive a bit deeper into the config you've written with another pass so apologies if this is figure-outable from a more careful review, but re: the port handling, did you wind up running into blockers with the bind-address flag OR is this something thats worked around in the current impl?

Overall, hugely appreciate the time you've spent on this and excited about the move to Inno Setup for a more modern windows installer.

build/windows/installer.iss
homandr: It's not a blocker per se, since shortcuts would still work. But if anyone selects a custom port and tries to use the tray menu, it will still go to 9898 and fail.

homandr:
The installer doesn't change Backrest behaviour regarding where the data is stored. It uses %appdata%\backrest regardless whether it's elevated or not. This is why the installer will refuse to install both non-admin and admin/user for the same user. It's technically possible to have both side by side through using batch files with BACKREST_CONFIG and BACKREST_DATA variables to have the secondary instance use a different location (and port needs to be different). But it's not something the installer supports. I doubt anyone would want to run two Backrest instances under the same user anyway.

When there are several users each with their own non-admin installation, the config is obviously unique for each, so no overlap there. When running under SYSTEM, it has it's own equivalent of %appdata%, which is %systemroot%\system32\config\systemprofile.

The admin with user context provides several benefits.
1. The main one—CIFS/SMB share access. Running under machine SYSTEM account loses all access to network resources unless it's wide open (not sure if it's even possible with Windows 11).
2. It starts Backrest only upon the specific user login. This could be beneficial for a shared PC with non-concurrent use. E.g. a user wants to run elevated but doesn't want to leave it running when the PC is used by someone else.
3. This might play a role for keys encryption in the future. In fact, I'm currently using PowerShell scripts to store repository and cloud tokens encrypted with Windows DPAPI, which leverages user account security to protect the data. Even with full admin access other users cannot gain access to encrypted data, and resetting the original user's password renders encrypted data unreadable.

build/windows/installer.iss
```diff
@@ -0,0 +1,352 @@
+#define B "Backrest"
+#define TrayExe "backrest-windows-tray.exe"
+#define Website "https://github.com/garethgeorge/backrest/"
+; The following is needed to extract the version from the change log.
+; If the application executable had the version info, then could use built-in GetVersion* functions.
```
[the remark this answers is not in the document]
garethgeorge: Hmm, golang actually has a weird approach to version embedding -- backrest already does something analogous to that snippet but what it's doing is setting a variable's value (by reflection iiuc) in the main package.

I'll take a bit of a look at this to see if we can simplify, but I'll take that as something for me to look at, your solution sgtm

garethgeorge:
Appreciate your expanding on that, I'll aim to followup with an updated README entry explaining the new installation options from that context-- overall makes a lot of sense and admin w/user context in that case solves a number of threads we've seen w.r.t. SMB share users on Windows.

I think this looks good to go ahead and merge now and target the next release.

garethgeorge:
Starting to work through release validations for the next minor version, figured out the port issue and fixed in https://github.com/garethgeorge/backrest/commit/da0ac093f19abf9ac17cc1ce310b8461ae8bf39c

maelstrm-code:
Thanks this is great! Btw please let me know if you'd prefer a new topic for these, I'm not familiar with the GitHub standard practice.

I wanted to share a few observations as a user who switched over to this:

1) When moving to the Admin > System User Context option the config file location changes. Launching Backrest then fails to find the config file so launches as a new install - might be good to add a check and an option to move or copy it during an install?

Manual fix was moving the contents of "C:\Users\MYUSERNAME\AppData\Roaming\backrest" across to "C:Windows\system32\config\systemprofile\AppData\Roaming\backrest"

2) After using Admin > System User, it's starting automatically (good) but seems like it has not run as admin, since VSS backup doesn't appear to be working (bad).  Task Manager shows the backrest.exe running as my current user.

3) The WebUI is showing [Version 1.7.2] even though I have 1.10.1 installed (to the best of my knowledge). Re-running the installer says that I'm running 1.10.1. Is there a way I can confirm it, and/or could this be right? The backrest.exe file in my has a last modified date of today (when I ran the installer last).

4) Installing with with Admin > Current User option resulted in Backrest's backend not being reachable even after multiple restarts. I can investigate a bit more if that would be helpful, though happy enough with Admin > System User if I can get that working.

homandr:
@maelstrm-code, the first item you mentioned is by design. You may want to switch over from one type to another, but other users may want to run both instances at the same time.

Number 2 shouldn't be happening, and I can't reproduce it either. I suggest you uninstall all instances and start from scratch. You already know where the configuration is stored, the installer doesn't touch it unless asked to delete during uninstallation.

The WebUI version text is not part of the installer code, so I can't comment on that. I can't reproduce the issue either.

Regarding *4, how do you access the UI? Make sure to use the shortcuts the installer creates. Admin mode uses port 9897, user mode uses 9898.

PR #899, opened 2025-09-13
fix: remove scheduled task time limit for Windows 2012

### LINKED ISSUE #897: Scheduled task for system-wide install configured to end after three days
Note: if you have a question or need support please post in the [discussions area](https://github.com/garethgeorge/backrest/discussions).

**Describe the bug**
When Backrest is installed using the newly added system-wide install feature, the scheduled task that is generated is configured to end itself after three days, resulting in Backrest stopping.

The task as generated is also configured to stop if the machine "ceases to be idle" or goes to battery power both of which could also be problematic.

**To Reproduce**

1. Install Backrest 1.9.2
2. Choose the system-wide install option
3. Wait three days
4. Backrest is no longer running, scheduled backups no longer occur.

**Expected behavior**
Backrest installed at the system level should remain running at all times and ideally restart itself if it crashes.

**Screenshots**

Image

**Platform Info**

- OS and Architecture: Windows Server 2012 R2 x64
- Backrest Version 1.9.2

homandr:
Fixes #897


garethgeorge:
Thanks for taking a look at this / the fix @homandr , this lgtm

PR #903, opened 2025-09-13
fix: merge windows tray functionality into backrest.exe for single binary install on windows

garethgeorge:


XhmikosR:
Hey, @garethgeorge. I just tested v1.10.0 and I don't see any tray icon at all on Windows 11. Backrest does run, but there's no tray icon. I use the Portable version like in v1.9.2.

garethgeorge:
Hmm, yeah this is weird. Looks like backrest parents itself to the shell's process if run from e.g. cmd.exe and is running, but doesn't start it's GUI

Please feel free to open a bug for that, I'll see if I can figure out why this regressed.

PR #905, opened 2025-09-16
feat: Clone existing plans

alyssaholland99:
Added a button to clone existing plans. This preserves everything about the plan you're trying to clone, while all fields remain editable, allowing you to use the same settings for different paths & repositories.

### DIFF
--- a/webui/src/views/AddPlanModal.tsx
+++ b/webui/src/views/AddPlanModal.tsx
@@ -65,18 +65,28 @@ const planDefaults = create(PlanSchema, {
   },
 });
 
-export const AddPlanModal = ({ template }: { template: Plan | null }) => {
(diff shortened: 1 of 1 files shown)


alyssaholland99:
<h2> Plan to clone, notice the new "Clone Plan" button. </h2>
Initial_Plan_To_Clone
<hr>



<h2>Modal shown after the 'Clone Plan' button is clicked, with the plan name changed manually</h2>
Cloned_Plan<hr>

<h2>Cloned plan saved successfully </h2>

Result_After_Submission

seb26:
Suggest change of button label from 'Clone Plan' to 'Duplicate'

Duplicate is used more widely than 'clone' in text editors, file browsers, etc, and clone has allusions to data cloning which this button's function is unrelated to

The other action buttons are also single word "Submit", "Delete" without the word plan

PR #906, opened 2025-09-17
feat: View plan details on hover

alyssaholland99:
Added a tooltip to show the plan details when the user hovers over them so that the user doesn't have to go into the settings to view what they're doing.

### DIFF
--- a/webui/src/views/App.tsx
+++ b/webui/src/views/App.tsx
@@ -9,8 +9,8 @@ import {
   LoadingOutlined,
 } from "@ant-design/icons";
 import type { MenuProps } from "antd";
-import { Button, Empty, Layout, Menu, Spin, theme } from "antd";
-import { Config } from "../../gen/ts/v1/config_pb";
+import { Button, Empty, Layout, Menu, Spin, theme, Tooltip } from "antd";
+import { Config, Schedule } from "../../gen/ts/v1/config_pb";
 import { useAlertApi } from "../components/Alerts";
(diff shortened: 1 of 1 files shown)


alyssaholland99:
Screenshot from 2025-09-17 15-48-53
Screenshot from 2025-09-17 15-48-47
Screenshot from 2025-09-17 15-48-42
Screenshot from 2025-09-17 15-48-37

garethgeorge:
Hi -- thanks for the interest in contributing, afraid for this one I have some UX / maintenance concerns.

introducing the tool tip means that JS logic to stringify the representation of the schedule must be maintained for every supported schedule type, I lean towards keeping one view of this implemented as the form UI.

The second issue I see is the paths included in the tooltip -- these aren't necessarily bounded to a small number, many users may have dozens or hundreds of paths included which would make for a _massive_ tool tip.

jburnham:
> The second issue I see is the paths included in the tooltip -- these aren't necessarily bounded to a small number, many users may have dozens or hundreds of paths included which would make for a _massive_ tool tip.

What about a `maxPathLines` and then an overflow `+ N more` line at the end. Sort of like this example
image

It's definitely a UX feature I've seen before, but of course when you need a live example you can't find one.

alyssaholland99:
Screenshot from 2025-10-07 10-30-37
Screenshot from 2025-10-07 10-29-08

alyssaholland99:
@garethgeorge Is this any better for limiting the paths to 6, with the +x more?

PR #907, opened 2025-09-18
feat: Clone existing repository config

alyssaholland99:
Added a button to clone existing repository configs. This preserves everything about the repository config you're trying to clone, while all fields remain editable, allowing you to use the same settings for different repositories.

### DIFF
--- a/webui/src/views/AddRepoModal.tsx
+++ b/webui/src/views/AddRepoModal.tsx
@@ -73,20 +73,32 @@ const repoDefaults = create(RepoSchema, {
   },
 });
 
-export const AddRepoModal = ({ template }: { template: Repo | null }) => {
+export const AddRepoModal = ({ template, templateClone }: { template: Repo | null, templateClone?: Repo | null }) => {
   const [confirmLoading, setConfirmLoading] = useState(false);
   const showModal = useShowModal();
   const alertsApi = useAlertApi()!;
   const [config, setConfig] = useConfig();
   const [form] = Form.useForm<JsonValue>();
-  useEffect(() => {
-    const initVal = template
-      ? toJson(RepoSchema, template, {
-          alwaysEmitImplicit: true,
-        })
-      : toJson(RepoSchema, repoDefaults, { alwaysEmitImplicit: true });
-    form.setFieldsValue(initVal);
-  }, [template]);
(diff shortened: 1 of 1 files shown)


alyssaholland99:
<h2> Repository config to clone, notice the new "Clone Repository Config" button. </h2>
Initial_Repository_To_Clone
<h2>Modal shown after the 'Clone Repository Config' button is clicked</h2>
Cloned_Repository_Config
<h2>Cloned repository config saved successfully </h2>
Cloned_Repository_Saved

alyssaholland99:
@garethgeorge The PRs I've put up should be ready for review if you have the time :)

garethgeorge:
Hi -- thanks for the interest in contributing, appreciate the PR! I want to start out with this is definitely a long standing issue and something I'd like to address, but I'm a bit opinionated about the right way to solve this.

I'm hesitant to introduce a new button on the submit rows of the repo / plan UIs since they'll become cluttered, duplicating plans or copying configuration is only something we think about when making a new plan or repo.

My thinking for a good solution to this is that it should be something like an optional dropdown when creating a new plan or repo (from an empty state) which allows the user to copy settings from an existing plan / repo. 

wdyt?

alyssaholland99:
Oh that's a good idea - I'll look at implementing this both on this PR and on the plan PR :)

seb26:
@garethgeorge

It's a more common UI trope to visit the 'one' you want to duplicate, to duplicate it

Of course 'Add Plan' | 'Add Plan from Existing' might be OK, but you suggest a dropdown or an arrow, that is now way more involved or prominent in the UI flow than a 'Duplicate' button

And if the 'Add Plan' blank window suggests to pull in config from another plan, then you have to have UI to permit selection of which plan, ... that's extra steps...

I think simplest most intuitive option wins out here...

> I'm hesitant to introduce a new button on the submit rows of the repo / plan UIs since they'll become cluttered

This can be addressed here by just using the button label "Duplicate" instead of the PR's suggestion "Clone Repository Config"

which unnecessarily repeats the modal context 'repository config', among neighbouring buttons which do not do this (e.g. 'Delete' not 'Delete Repository') 

There are not many other possible buttons to add here, either, so the clutter isn't already happening nor likely to

PR #926, opened 2025-09-29
fix: Avoid shared cache memory databases

ncruces:
My driver does not support [shared cache](https://github.com/ncruces/go-sqlite3/discussions/97), so you'll eventually bump into this issue: https://github.com/ncruces/go-sqlite3/discussions/188

The preferred pattern for tests is to use [`memdb.TestDB`](https://pkg.go.dev/github.com/ncruces/go-sqlite3/vfs/memdb#TestDB); this handles test cleanup as well.

### DIFF
--- a/internal/api/syncapi/peerstate_test.go
+++ b/internal/api/syncapi/peerstate_test.go
@@ -5,9 +5,9 @@ import (
 	"testing"
 	"time"
 
-	"github.com/garethgeorge/backrest/internal/cryptoutil"
 	"github.com/google/go-cmp/cmp"
 	_ "github.com/ncruces/go-sqlite3/driver"
+	"github.com/ncruces/go-sqlite3/vfs/memdb"
 )
 
 func PeerStateManagersForTest(t testing.TB) map[string]PeerStateManager {
@@ -92,7 +92,7 @@ func TestPeerStateManager_OnStateChanged(t *testing.T) {
 
 func newDbForTest(t testing.TB) *sql.DB {
 	t.Helper()
-	dbpool, err := sql.Open("sqlite3", "file:"+cryptoutil.MustRandomID(64)+"?mode=memory&cache=shared")
+	dbpool, err := sql.Open("sqlite3", memdb.TestDB(t))
 	if err != nil {
 		t.Fatalf("error creating sqlite pool: %s", err)
 	}
--- a/internal/api/syncapi/syncapi_test.go
+++ b/internal/api/syncapi/syncapi_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/garethgeorge/backrest/internal/resticinstaller"
 	"github.com/garethgeorge/backrest/internal/testutil"
 	"github.com/google/go-cmp/cmp"
+	"github.com/ncruces/go-sqlite3/vfs/memdb"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
 	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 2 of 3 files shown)


garethgeorge:
Exciting you noticed the migration, thanks for catching that & appreciate your fixing my mistake in tests.

Going ahead and merging.

As a side note-- some interesting benchmark numbers comparing your lib to zombiezen in the context of backrest:

```
goos: darwin
goarch: arm64
pkg: github.com/garethgeorge/backrest/internal/oplog/storetests
cpu: Apple M4 Pro
BenchmarkAdd/memory-12           4855030              1173 ns/op
BenchmarkAdd/sqlite-12              4454            314339 ns/op
BenchmarkAdd/sqlitewal-12          21926             56832 ns/op
BenchmarkAdd/zensqlite-12          16341             75813 ns/op
BenchmarkList/100/memory-12                48213             24493 ns/op
BenchmarkList/100/sqlite-12                15726             78142 ns/op
BenchmarkList/100/sqlitewal-12             15511             83535 ns/op
BenchmarkList/100/zensqlite-12             18264             64482 ns/op
BenchmarkList/1000/memory-12                4293            284047 ns/op
BenchmarkList/1000/sqlite-12                1980            602186 ns/op
BenchmarkList/1000/sqlitewal-12             2019            584893 ns/op
BenchmarkList/1000/zensqlite-12             2589            451032 ns/op
BenchmarkList/10000/memory-12                405           2916554 ns/op
BenchmarkList/10000/sqlite-12                200           5758259 ns/op
BenchmarkList/10000/sqlitewal-12             207           5721702 ns/op
BenchmarkList/10000/zensqlite-12             262           4528795 ns/op
BenchmarkGetLastItem/100/memory-12        403381              2748 ns/op
BenchmarkGetLastItem/100/sqlite-12         59106             19642 ns/op
BenchmarkGetLastItem/100/sqlitewal-12              72483             16579 ns/op
BenchmarkGetLastItem/100/zensqlite-12              83940             14341 ns/op
BenchmarkGetLastItem/1000/memory-12                24061             49078 ns/op
BenchmarkGetLastItem/1000/sqlite-12                60926             19340 ns/op
BenchmarkGetLastItem/1000/sqlitewal-12             70303             16579 ns/op
BenchmarkGetLastItem/1000/zensqlite-12             83389             14538 ns/op
BenchmarkGetLastItem/10000/sqlite-12               61255             19330 ns/op
BenchmarkGetLastItem/10000/sqlitewal-12            72597             16885 ns/op
BenchmarkGetLastItem/10000/zensqlite-12            82284             14566 ns/op
BenchmarkGetLastItem/10000/memory-12                1900            620708 ns/op
PASS
ok      github.com/garethgeorge/backrest/internal/oplog/storetests      59.974s
```

In general, pretty happy with the performance on par or better (in the case of writes) than zombiezen considering that the driver is going through a WASM layer-- which is also what drove me to switch for ease of cross compilation + simpler dep management than zombiezen.

Looks like WAL mode enabled write performance is vastly better with your driver, I'll likely go ahead and commit something to enable that by default. Can't quite tell from your docs what the status of WAL is / whether it's robust? Would you recommend it?

ncruces:
WAL should be fine in all the platforms where "Shm WAL" is green in this [table](https://github.com/ncruces/go-sqlite3/wiki/Support-matrix). The big exception is Windows, where you should have somewhat reduced performance. In general, green and yellow in that table should be OK; anything red or orange you should avoid if possible.

As for performance, this seems consistent with `fsync` being slow on macs, which is a big topic to cover. You should read a bit on [`PRAGMA synchronous`](https://sqlite.org/pragma.html#pragma_synchronous).

For my driver, and default for rollback is FULL, and for WAL is NORMAL, as [recommended by SQLite](https://sqlite.org/compile.html#recommended_compile_time_options); zombiezen might be different.

On a mac should also decide if you want [`PRAGMA fullfsync`](https://sqlite.org/pragma.html#pragma_fullfsync) or [`checkpoint_fullfsync`](https://www.sqlite.org/pragma.html#pragma_checkpoint_fullfsync).

PR #927, opened 2025-09-29
feat: Add restic and OS details to system info

inode64:
When managing multiple systems, it's important to know where I'm connected, the version of backrest, and the OS.
I've included the uptime information because I've occasionally experienced issues with the server rebooting. This way, I know how long backrest has been active.

And it has also happened to me that the DataPath directory grows excessively and fills all the space, so you can control how much it is being used


internal/api/backresthandler.go
```diff
@@ -909,6 +911,21 @@ func (s *BackrestHandler) GetSummaryDashboard(ctx context.Context, req *connect.
 	return connect.NewResponse(response), nil
 }
 
+func humanTime(d time.Duration) string {
```
garethgeorge, line 917: super nit: let's return the duration as a uptimeMillis (int64 milliseconds is the common time resolution used in most places in backrest), this also makes it more useful for scripting.

We can reuse formatting logic already implemented in js in the webui, there's a utility for this is available in https://github.com/garethgeorge/backrest/blob/main/webui/src/lib/formatting.ts#L91-L118

internal/api/backresthandler.go
```diff
@@ -886,3 +913,82 @@ func (s *BackrestHandler) GetSummaryDashboard(ctx context.Context, req *connect.
...
+	if size < unit*unit*unit {
+		return fmt.Sprintf("%.2f MB", float64(size)/(unit*unit))
+	}
+	if size < unit*unit*unit*unit {
+		return fmt.Sprintf("%.2f GB", float64(size)/(unit*unit*unit))
+	}
+	return fmt.Sprintf("%.2f TB", float64(size)/(unit*unit*unit*unit))
+}
+
+func directorySize(path string) string {
```
garethgeorge, line 949: Hmm, I'm a bit nervous about the performance of this function. In principle checking the size is a neat idea, in practice it's likely to be the slowest part of loading the status summary dashboard as it's implemented right now.

My inclination is to say it should be dropped, or if we report something it should be the sizes of specific stores e.g. # total operations, # total operation log bytes used, etc. Same goes for logs, but I'm not sure I want to expose info at that level of granularity.

garethgeorge:
Broadly looks great, thanks for the interest in contributing the extra details. Left a few nits. Only summary detail I have some reservations about is the storage used mostly from a performance perspective.

Overall appreciate the extra details -- agree that this will help users with differentiating installs and keeping track of uptime for noticing crashes (which I _think_ are very rare) is a nice addition.

inode64:
Yes, I understand the problem this could pose.
Also, now that I'm not really seeing the serial dataDir, what worries me most is the rapid growth. I see that it's between 8 and 12 MB.
The restic cache, which in some cases exceeds 4 GB.
I'm thinking of something more optimised; otherwise, I'll deactivate it.

garethgeorge:
Hmm, I was thinking on this a bit longer -- we can probably leverage the summary dashboard's refresh logic for a decent solution, see https://github.com/garethgeorge/backrest/blob/3ccb883b2f47880017dcb0dd3383cfdf58836929/webui/src/views/SummaryDashboard.tsx#L78

What if we do the following 

 1. Add an additional once off refresh after ~5s on the first load of the page.
 2. On the first request to the summary dashboard trigger updating slow properties e.g. the restic version, data directory size, etc in the background. Compute these once and cache them (either forever, or to be refreshed only after some TTL). But return immediately w/a partial response if they aren't ready in the cache.

In this way, the first request will load fast, the second request follows up with slow to compute data.

internal/api/backresthandler.go
inode64: Adjusted to use the function

inode64:
I've removed the space calculation as it can be a performance penalty.
I'll think about the best way to do it for another pull request.

internal/api/backresthandler.go
```diff
@@ -858,6 +859,29 @@ func (s *BackrestHandler) GetSummaryDashboard(ctx context.Context, req *connect.
...
+		response.Hostname = hostname
+	}
+
+	if version := systemOSVersion(); version != "" {
+		response.OsVersion = version
+	}
+
+	if s.orchestrator != nil {
+		response.ResticPath = s.orchestrator.ResticBinary()
+		if response.ResticPath != "" {
```
garethgeorge, line 874: This one is quite slow as well-- let's defer it to the future PR as well, with that change I think this LGTM to merge!

garethgeorge:
Hi, closing this out as inactive for now. Please revive if still interested!

PR #928, opened 2025-09-29
fix: Improve JSON parsing resilience

inode64:


garethgeorge:
Thanks for contributing this, does a good job of cleaning up the logic.

Out of curiosity -- was there a functional issue that this addresses? From my review, looks like it should be functionally equivalent, but improves code clarity. Or is there a behavioral change as well that I'm missing?

inode64:
It was for simplicity, but if the test works now without applying the changes, we can leave it.
If something works, sometimes it's better not to touch it; if it doesn't, it improves clarity, of course.

garethgeorge:
Happy to go ahead and merge it -- I feel it clarifies the implementation & the new test coverage is beneficial.

Issue #946, opened 2025-10-11
Add openrc support


AlbertoBoc:
Note: if you have a question or want discussion please post in the [discussions area](https://github.com/garethgeorge/backrest/discussions).

**Is your feature request related to a problem? Please describe.**
I use a non systemd system, I'd like to add support for openrc.

**Describe the solution you'd like**
My solution is to edit the install.sh and uninstall.sh script to:
1. Detect whether running on a systemd or openrc system. *See additional context*
2. Create a systemd unit or an openrc service script depending on the context
3. [uninstall.sh] Delete the systemd unit or the openrc service script depending on the context

**Additional context**
I'd like to point out that checking whether _/etc/systemd/system_ exists is NOT a robust detection method. I know by personal experience: another software created _/etc/systemd/system_ at installation time, disregarding the fact that I run openrc. 
Because of this, please do notice that in my pull-request you'll find a detection method based on the existance of systemctl and rc-status, which I consider to be robust.

PR #947, opened 2025-10-11
Feat: adds the installation of an openrc service script

AlbertoBoc:
This is my own implementation, responding to the Issue #946 (Add openrc support) which I myself opened.

I'd like to point out again that I did change the systemd (now systemd/openrc) detection method.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -28,12 +28,16 @@ else
   echo "Local access only: Backrest will bind to 127.0.0.1:9898, run ./install.sh --allow-remote-access to enable remote access"
 fi
 
-install_or_update_unix() {
+stop_systemd_service() {
   if systemctl is-active --quiet backrest; then
     sudo systemctl stop backrest
     echo "Paused backrest for update"
   fi
-  install_unix
+}
+
+stop_openrc_service() {
+  sudo rc-service backrest --ifstarted stop
+  echo "Paused backrest for update (if started)"
 }
 
 install_unix() {
@@ -44,11 +48,6 @@ install_unix() {
 }
 
 create_systemd_service() {
-  if [ ! -d /etc/systemd/system ]; then
-    echo "Systemd not found. This script is only for systemd based systems."
-    exit 1
-  fi
-
   if [ -f /etc/systemd/system/backrest.service ]; then
     echo "Systemd unit already exists. Skipping creation."
     return 0
@@ -74,6 +73,41 @@ EOM
 
   echo "Reloading systemd daemon"
   sudo systemctl daemon-reload
+
+  echo "Enabling systemd service backrest.service"
+  sudo systemctl enable backrest
+  sudo systemctl start backrest
+}
+
+create_openrc_service() {
+  if [ -f /etc/init.d/backrest ]; then
+    echo "Openrc service already exists. Skipping creation."
+    return 0
+  fi
+
+  echo "Creating openrc service at /etc/init.d/backrest"
+
+  sudo tee /etc/init.d/backrest > /dev/null <<- EOM
+#!/sbin/openrc-run
+description="Backrest Service"
+
+depend() {
+    need loopback
+    use net logger
+}
+
+command=/usr/local/bin/backrest
+command_background=true
+pidfile="/run/${RC_SVCNAME}.pid"
+
+export BACKREST_PORT=$BACKREST_PORT
+EOM
+
+  sudo chmod 755 /etc/init.d/backrest
+  echo "Adding backrest to runlevel default"
+  sudo rc-update add backrest default
+  echo "Reloading openrc service"
+  sudo rc-service backrest start
 }
 
 create_launchd_plist() {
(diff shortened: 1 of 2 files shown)


AlbertoBoc:
Sorry, there were two bugs that have now being fixed.
The most important part is that my service was running as root, now it is running as user, I had forgotten to translate that very important line from the systemd unit.

Actually, I'd rather have backrest running with the new openrc "User services" mechanism, but so far I didn't get much luck, so I'll just provide this one-2-one translation of your systemd unit.

garethgeorge:
Thanks for the contribution, I think this change seems reasonable and I'm happy enough to support it in the installer.

The biggest caveat I'd call out is that I don't have a system with openrc or any ability to test this on future upgrades or changes to the script. I'm inclined to err on the side of merging this since I don't forsee a reason it'd _just stop working_ or that i'd want to overhaul the installer, but it's very possible that this installer capability could get broken (occasionally) in the future.

AlbertoBoc:
Hello, I 100% understand your point. I'm willing to maintain this feature going forward. I'm using your program (of course!), so I'll update it as new versions come and the fact that's an active development project with many PRs, open issues etc. makes me believe that's worthed to support it.
If you decide to merge it, I'll install the new versions on both openrc and systemd systems, so I can take care of fixing errors. For your information (in the age of containers and VMs maybe you're willing to "virtually try it"), openrc distros are Alpine, Artix and Gentoo. Distrowatch has got many more, I listed the ones who come to mind!

AlbertoBoc:
Sorry, I had forgotten to sign my commits and had my good time repairing that.

inode64:
This is my openrc for gentoo

<pre>
#!/sbin/openrc-run

name="backrest"
description="backrest Daemon"
command="/usr/bin/backrest"
command_args="-restic-cmd /usr/sbin/restic -data-dir /var/lib/backrest/ -config-file /var/lib/backrest/config.json"
command_background=true
pidfile="/run/backrest.pid"
supervisor=supervise-daemon

output_log="/var/log/backrest/backrest.log"
error_log="/var/log/backrest/backrest.log"

depend() {
    need net
}
</pre>

from:
https://github.com/inode64/inode64-overlay/tree/main/app-backup/backrest

I think it would not be necessary to require a logger

AlbertoBoc:
Hello,

Thank you for chiming in. First of all, I'd like to know why you add all those args. I felt they were actually NOT needed, I only exported BACKREST_PORT as a variable (not elegant, I know.), but  I'm totally in for including them if they are needed. And I'm particularly referring to the restic part. My understanding was that the current restic installation is automatically detected by backrest via a "restic" command call, but if I'm wrong it is clear that we must use your version of that line, instead.

I don't "need" a logger, but I "use" a logger and the network ("if present, please start them before backrest"). On the logger side, I might agree with you, since backrest logs on his own. 
On your line "need net", I don't agree, but I'm open to change my mind. My feeling at writing time was that we "need" loopback, but we "use" net (feel free to disagree, maybe we "want" or really "need" it?). In fact, in my case I bind to 127.0.0.1, I need loopback. There might be cases were the user wants network, but they cannot have it for modules/drivers issues. Backrest in that case might still run on loopback. Then we can discuss the usefulness of local backups (no rclone, no S3...), but that's another story...

I see that your script misses the user line, are you running it as root. If so, why? Once I used it like this, all the backups were root 700 (if don't go wrong). Why would root own my backups?

Finally, you put in the line for "supervise-daemon", I've been on the edge of putting it into my script, too, but then I simply didn't. Do you think that's a critical service that must/should be supervised? Maybe we should add it?

Finally, on an Off Topic [sorry], but definitely interesting, note... Have you got an ebuild for this? EDIT: sorry, I just investigated your link better, THAT IS an overlay. I was more of the idea of creating an overlay with ONLY backrest, but I see you've got many things there... I'll check that out.

Thank you for your contribution,
Alberto

garethgeorge:
> On your line "need net", I don't agree, but I'm open to change my mind. My feeling at writing time was that we "need" loopback, but we "use" net (feel free to disagree, maybe we "want" or really "need" it?). In fact, in my case I bind to 127.0.0.1, I need loopback. There might be cases were the user wants network, but they cannot have it for modules/drivers issues. Backrest in that case might still run on loopback. Then we can discuss the usefulness of local backups (no rclone, no S3...), but that's another story...

I think skipping need net is fine, assuming this waits for LAN / WAN? The right way to handle this imo is to leave it up to the user to define backup start healthchecks that await the network.

> I see that your script misses the user line, are you running it as root. If so, why? Once I used it like this, all the backups were root 700 (if don't go wrong). Why would root own my backups?

right, I think whoami is the best way to do this. Users wanting to use root can run the installer with sudo.

> signing

thanks for taking care of commit signing, overall this lgtm. Will go ahead and merge as is for the next release. Thanks for contributing the support

inode64:
I use the root user because I use it as a backup for any data on the servers, email, web, SQL, etc., and I don't have restrictions when copying. Of course, this can have security implications, but in my case, I haven't seen a more convenient solution for backups.
Then I force the restic location to use the one installed on the system.

NicholasFlamy:
As somebody who doesn't write init/service files, I looked up whether this was compatible with sysvinit (without openrc) (the default for Devuan) and that seems to be the case so let's hope it works.

AlbertoBoc:
How are things going? According to my information it is possible to use sysvinit scripts with openrc, but not the other way around. Is it working for you?

NicholasFlamy:
> How are things going? According to my information it is possible to use sysvinit scripts with openrc, but not the other way around. Is it working for you?

Yep, what you said sounds right. I temporarily made the install script install openrc's init script for sysvinit and then used AI to adapt the init script (since it's a very basic init script). It seems to work fine.

I'll be making a PR to add sysvinit support soon enough. (And don't worry, I'll be sure to clean things up.)

PR #950, opened 2025-10-16
Use default locale for formatting dates

michaelpearson:
Removed hard-coded locale from `Intl.DateTimeFormat`. Ensures that dates are formatted in the users locale.

It's not entirely clear to my why `US` was picked so I could be missing something!


garethgeorge:
Very silly of me -- I intended for this to be internationalized, good catch and thanks for contributing the fix. LGTM.

PR #951, opened 2025-10-16
fix: getOperationEvents use create EmptySchema not javascript empty object

ASNFalcon:
backrestService.getOperationEvents parameters change to use create(EmptySchema)  to avoid the request's body from webui have 5 NULL content and calls the backend  504 Timeout, 
image


garethgeorge:
Lgtm, thanks for the bug fix.

PR #952, opened 2025-10-17
Add podman to Alpine image

ptomli:
Useful to allow managing of container with podman before/after exporting volumes. Unfortunately does add a few MB to the image.


garethgeorge:
Hey -- thanks for the interest in contributing, but I'm afraid I've decided to freeze the list of dependencies added to the alpine image to the minimal set I have now.

I want to support basic scripting, but recommend users build their own images off of the alpine image for any additional cli tools they may need. I want to bound the growth of the image & the need to worry about dependency updates when publishing new images.

PR #975, opened 2025-10-31
fix: overwrite service scripts - supervise openrc daemon

AlbertoBoc:
The two commits are kinda self-explanatory.
During my last pull request, I submitted a new install script. I just realised that my version did not start back the service after terminating it, in the case in which a service script/systemd unit was already installed.
Aside from my bug, also there is a point in updating the service scripts together with the software in case it is needed. The simplest case is the case in which the user forgets _--allow-remote-access_ and wants to reinstall with a different port binding. This fixes the problem.

As it can be read in https://github.com/garethgeorge/backrest/pull/947 it was the user @inode64 that suggested to add the supervised daemon line in the openrc script. I don't think this change is big enough to justify a pull request of its own, forgive me for including it here.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -65,6 +65,7 @@ Type=simple
 User=$(whoami)
 Group=$(whoami)
 ExecStart=/usr/local/bin/backrest
+Restart=on-failure
 Environment="BACKREST_PORT=$BACKREST_PORT"
 
 [Install]


garethgeorge:
Hmm, I've been deliberately avoiding changes that would require the default sets of env vars, etc in the systemd units to be updated. I'm not sure about changing to always always overwriting them.

I suspect many users may have customized ports, etc in their systemd units at this point. 

Is the main fix here the line adding the supervisor daemon that will restart backrest on crash to the openrc unit?

AlbertoBoc:
I'd describe supervising the daemon as an additional layer of security rather than a fix... I've been using this for three weeks now, cannot remember backrest crashing... you never know.

Anyway I see where you're coming from. I agree with you that, given the absence of a config file, people will go crazy in those scripts, instead. To this point, are you looking to add an /etc/backrest config? That'd be nice.

I reverted the commit and I restart the stopped daemon in a different part of the code. I'd say that the daemon not restarting after installation is the main point if this  pr.

garethgeorge:
> I'd describe supervising the daemon as an additional layer of security rather than a fix... I've been using this for three weeks now, cannot remember backrest crashing... you never know.

I've definitely taken pains that it (hopefully) shouldn't ever be crashing. I haven't seen many reports of it ... so yeah lack of automatic restarts probably won't be noticed by many users since restarts are very infrequent. But it's backup software -- I'm glad you're taking care re: any case where it can just stop running silently in the background and go unnoticed.

> Anyway I see where you're coming from. I agree with you that, given the absence of a config file, people will go crazy in those scripts, instead. To this point, are you looking to add an /etc/backrest config? That'd be nice.

Hmm, no plans on an /etc/backrest type config, backrest tries to locate its configs around the user account it's installed / running as so always in ~/.local/share for data and ~/.config for configs.

> I reverted the commit and I restart the stopped daemon in a different part of the code. I'd say that the daemon not restarting after installation is the main point if this pr.

Yep, adding the code to start it up is a good catch, thanks.

garethgeorge (approved):

PR #1006, opened 2025-11-25
fix: FormatSizeBytes reports incorrect sizes with fractional-gigabytes.

clnhlzmn:
I have noticed the size reporting in the .Summary template function being incorrect. The issue is the value on both sides of the decimal place is the same.

For example:

image

Now.. I'm not yet sure if this change is how the formatting should be done either.


garethgeorge:
Thanks for taking a pass at this and contributing a fix, LGTM. Going ahead and merging.

PR #1031, opened 2025-12-23
fix: it.json - Italian translation improvements

ziotibia81:
Key Italian translation improvements.
It requires a more surgical approach.


garethgeorge:
Thank you for contributing this, really love to see that the translations are valuable and appreciate the time you've taken to touch these up even before the first release with translations.

I'm going to have to take your word for the grammar improvements :) lgtm.

AlbertoBoc:
I confirm that this one is more accurate than the previous.
I only have doubts about removing the verb "potare" everywhere in favour of "pulire". In the context of data structures, most obviously with trees, the word "potare" is in use in Italian, too. It is difficult for me to decide whether I would prefer potare or pulire, out of context. Probably pulire, I won't lie, since it is more adequate in most contexts!
Aside from this little note, the word choice and grammar is definitely the right one now.

ziotibia81:
I have never encountered the term _'potare'_ outside of an agricultural context. _'Pulizia backup'_ is the standard term. Alternatively, we could retain the English term _'prune'_, since _'potare'_ is unsuitable in this context.

AlbertoBoc:
I found the place where the word is used, I agree that "pulire" is probably the best guess, as a non-technical speaker would probably not understand "potare". In the context of tree data structures "potare un ramo" (pruning a branch) means "deleting a node with all of his children and so on recursively". 
In view of the fact that the variable "repo_tooltip_prune" explains what pulire means in this context, I guess that this is the best choice.

PR #1032, opened 2025-12-24
fix: zh.json - Simplified Chinese translation

yhdsl:
Hi, thank you for your continued contributions to this project.

Previously, I personally maintained a Simplified Chinese fork. I’m very excited to see that you’ve now added i18n support to the project. Based on my fork, I recreated the zh.json file and polished the previous translation texts.

*Unfortunately, there is still a large amount of text that has not yet been included in the i18n system. I will wait until the related work is completed before continuing to update the zh.json file.*


yhdsl:
*Additionally, when I was creating the Simplified Chinese fork, I encountered several other i18n-related issues that require extra attention.*

The first one is that the Cron element in the react-js-cron component requires manual configuration of translation texts. This commit shows a Simplified Chinese implementation: https://github.com/yhdsl/backrest/commit/88167517f5f0204639045dc62783ce219523a3d0

yhdsl:
Secondly, in `RunCommandModal.tsx`, the run command popover displays "No Data" when there is no data available by default. This text comes from an Ant Design component and requires configuring the Ant Design locale in order to be localized properly.

A fix example for the Simplified Chinese version:
```
import zhCN from 'antd/locale/zh_CN';

  return (
      <ConfigProvider locale={zhCN}>
        <Modal
        ...
        </Modal>
      </ConfigProvider>
  );
};
```

Related commits (may be a bit lengthy): https://github.com/yhdsl/backrest/commit/9b7c4b068685d268ff972ebcd3afe1911806efb6#diff-8c953d5e262a3d8cc4966489573f22c316d380d3998b84b9eedf8a9d9501cee7

yhdsl:
Finally, in `SnapshotBrowser.tsx`, when backup file names contain non-English characters, the current sorting behavior becomes somewhat inconsistent. It would be preferable to use *natural sorting* in this case.

---

Thank you again for your generous and selfless contributions. If needed, please feel free to split these into multiple sub-issues.

garethgeorge:
Brilliant, really exciting first off to hear you were motivated to fork it for simplified chinese support in the first place and glad I can make life easier with upstream translation support :)

Translations were on my roadmap for a while, but obviously add to my maintenance cost so I'd been holding off, seeing the immediate interest really validates the work it takes to support them, I appreciate it.

I'll get this merged in the next few days, and +1 re: holding off on the full set of strings. Also keep your eye out for https://github.com/garethgeorge/backrest/pull/1034 . I have a UI overhaul in the works, part of that project will be ensuring the rest of the text in the UI is available as translation contants.

garethgeorge:
> Finally, in `SnapshotBrowser.tsx`, when backup file names contain non-English characters, the current sorting behavior becomes somewhat inconsistent. It would be preferable to use _natural sorting_ in this case.

Re: this issue, please file a bug for this. Is there a library you're familiar with that handles this correctly for Simplified Chinese / other unicode character sets? Is using something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare sufficient?

yhdsl:
> Re: this issue, please file a bug for this. Is there a library you're familiar with that handles this correctly for Simplified Chinese / other unicode character sets? Is using something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare sufficient?

OK, I’ve created https://github.com/garethgeorge/backrest/issues/1035 and provided a working fix.

yhdsl:
Additionally, I have a small translation-related request: could you add placeholders for the numbers at the beginning of these texts? The construction of numbers + nouns feels awkward in the translation.

```
  "dashboard_card_backups_ok": "ok",
  "dashboard_card_backups_failed": "failed",
  "dashboard_card_backups_warning": "warning",
```

PR #1055, opened 2026-01-12
fix: fr.json - French translation

Lynilia:
Thank you for adding localization support to Backrest!
Machine translation for French is often quite bad, producing some completely unintelligible strings, so I reworked most of it.
While it might still need some tweaks (which will be easier to notice once released), it should provide a solid base.


garethgeorge:
Thank you for contributing this, really appreciate the engagement on the translations. Going ahead and merging.

PR #1076, opened 2026-01-19
fix: minor translation errors (de/german)

derphilipp:


Speaka:
"clients" should not be translated with "Kunden" as it is a technical term also in german.

garethgeorge:
Hi, thanks for contributing the translation fixes. Greatly appreciated. Realizing I merged another German translation patch before getting to this one and am seeing merge conflicts. If you have a bit of bandwidth for a quick merge conflict pass it'll be much appreciated (realizing i'm struggling to know which branch is correct in most cases). With that, happy to go ahead and merge this.

PR #1081, opened 2026-01-22
fix: update German translations for backrest messages

Speaka:
Changed several strings because 'backrest' was translated literally; it should remain as a brand name. Also updated the help messages, as they are prominent in the GUI.


garethgeorge:
Awkward indeed, thanks for catching this and contributing corrections.

PR #1094, opened 2026-01-27
chore: update troubleshooting SFTP documentation

wirehiker:
Improved troubleshooting methods: verbose ssh, user known_hosts instead of global while testing connection from host.


garethgeorge:
Thanks for contributing this -- much needed, I think all of the suggestions you've made are good ideas.

Also want to link https://github.com/garethgeorge/backrest/pull/961 , next release will include streamlined SFTP setup which will hopefully go a bit further towards encouraging SFTP adoption as well.

wirehiker:
Thanks for pointing out #961. I checked out the conversation and "contributed" with a couple of thoughts.

Issue #1105, opened 2026-02-04
backrest.service is disabled after update


OliverSCE:
Im an running backrest under proxmox in an ubuntu-server container.

Since several releases I have the issue that after updating backrest the service is disabled, and needs to be enabled manually.

* backrest.service - Backrest Service
     Loaded: loaded (/etc/systemd/system/backrest.service; disabled; preset: enabled)

I update by downloading the latest *.tar.gz file, unpack it, and run the install.sh script.

AlbertoBoc:
I created this error, I thought I had also solved it, but clearly I missed something.
Please bear with me as I work on this issue in the next few days. Sorry for the trouble.

AlbertoBoc:
I hope pr #1113 fixes the problem.
Please do not close this issue until the solution is confirmed by the user.
Thank you

OliverSCE:
Problem is fixed. Thank you.

PR #1113, opened 2026-02-13
fix: enable systemd unit, change rc service

AlbertoBoc:
This SHOULD solve issue #1105 

I also updated the open rc-service.


garethgeorge:
Looks great-- thank you for contributing a fix here.

PR #1131, opened 2026-02-22
fix: use bufio.Reader for reading restic ls --json

jeffychien:
`restic ls --json` on snapshots created with `restic --files-from-verbatim` lists all paths in the snapshot info line (first line), which can easily exceed the default 64KB token limit of `bufio.Scanner`. This changes it to `bufio.Reader` which reads indefinitely until the token.

Tested on repos that are causing the issue.

### DIFF
--- a/pkg/restic/outputs.go
+++ b/pkg/restic/outputs.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"time"
+	"strings"
 
 	v1 "github.com/garethgeorge/backrest/gen/go/v1"
 )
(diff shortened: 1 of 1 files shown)


garethgeorge:
Hi, thanks for contributing this patch -- just went ahead and kicked off a presubmit run but it looks like a few tests are failing related to ReadLs. 

Take a look at https://github.com/garethgeorge/backrest/actions/runs/22276993763/job/64462605442?pr=1131 

I wonder if `reader.ReadBytes('\n')` exactly matches the prior behavior e.g. if there's an unterminated line at the end of the output stream?

jeffychien:
Hi, thanks for the quick feedback, and sorry about that! It should pass now, and I also booted it up and checked that it listed all files from a smaller test repo.

garethgeorge:
Hi, appreciate the test fixes! Thanks for taking a look at those -- left a few additional comments, realizing the new line handling is raising some concerns. I'm not sure how restic handles line terminators across platforms, and off the top of my head I'm thinking ideally we can delegate handling that to the golang standard library.

Wdyt re: trying to simply bump up the buffer size e.g. here https://pkg.go.dev/bufio#Scanner.Buffer , looks like the scanner periodically expands the buffer until it hits the configured limit. I think we can set this to a reasonably large fixed number e.g. 64MB or 128MB. Hesitant to go much larger than that.

jeffychien:
Thanks for the response. I'm not much of a go expert (hence the initial errors) so I would be a little uncomfortable implementing that myself. I think upstream is guaranteed to only send newlines instead of some mix with carriage returns: https://github.com/restic/restic/blob/master/cmd/restic/cmd_ls.go

But if you insist on the scanner/retry way I think you should implement it. I hope the user is able to set the initial buffer size.

I do think there is an opportunity to improve restic because (1) `--files-from*` has an unexpected side effect and (2) as you say we're unable to workaround that side effect without dropping the first line manually.

garethgeorge:
Put in https://github.com/garethgeorge/backrest/commit/bebe60850566c603eb37659daf89b853ffdc9f63 to bump the buffer size, let's try this and see if it fixes the problem you were running into.

PR #1133, opened 2026-02-22
Update api.md

AdmiralSprk:
Updated API documentation to include sequential backup triggering. Rephrased password authentication so it reads like authentication it is the norm, not the exception. Included a note about special characters, as this was a stumbling block for me.


garethgeorge:
Thank you for the doc contribution, the revisions look great. Going ahead and merging.

Issue #1150, opened 2026-03-02
Release for 32 bit armv7


csm10495:
**Is your feature request related to a problem? Please describe.**
I was trying to use backrest to backup some directories on an older raspberry pi via docker. Wasn't able to because of this error:

```
z:\>docker --context pi1 run -it --rm garethgeorge/backrest:v1.12.0 sh
Unable to find image 'garethgeorge/backrest:v1.12.0' locally
v1.12.0: Pulling from garethgeorge/backrest
docker: no matching manifest for linux/arm/v7 in the manifest list entries

Run 'docker run --help' for more information
```

**Describe the solution you'd like**
The image to be released for the existing architectures and for linux/arm/v7 (so i can use it with my pi)

**Additional context**
Much thanks for considerations. I don't think it would be too hard to do this, though honestly haven't dug into it too much.

garethgeorge:
This is not implemented, marking as resolved :)

csm10495:
Hey @garethgeorge can you reopen? Maybe I missed something but on https://hub.docker.com/r/garethgeorge/backrest/tags i only see arm/v6 for the scratch images, but not the 'regular' or non-scratch ones.

PR #1165, opened 2026-03-12
Add Alpine Docker images for arm/v6 and arm/v7 architectures

csm10495:
Add some missing docker hub arch refs. 

Related to https://github.com/garethgeorge/backrest/issues/1150
and comment: https://github.com/garethgeorge/backrest/issues/1150#issuecomment-4050803522

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -116,6 +116,36 @@ dockers:
       - docker-entrypoint
 
   - image_templates:
+      - garethgeorge/backrest:{{ .Tag }}-alpine-armv6
+      - ghcr.io/garethgeorge/backrest:{{ .Tag }}-alpine-armv6
+    dockerfile: Dockerfile.alpine
+    use: buildx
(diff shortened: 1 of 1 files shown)


csm10495:
@garethgeorge mind checking when you get to it?

garethgeorge:
Hi, apologies for the late review -- somewhat hesitant re: the alpine and armv6 alpine versions. This commits to dependencies that are available on those architectures in the arm image (and adds more cost to the release pipeline).

On armv6 / armv7 devices are you finding yourself using scripting features e.g. docker in the container etc for hooks?

csm10495:
On my old raspberry pi, I need docker inside the image so I can pause a DB container during backup. 
So pre-hook: stop db containers
Post-hook: restart db containers.

garethgeorge:
Thanks for explaining -- I think we can go ahead with this for raspberry pi devices. Probably a common enough usecase that it's worth the build time.

PR #1199, opened 2026-04-18
Build for i386 systems

pwbriggs:
Hey! I'm trying to run Backrest on an old 32-bit server box. Would you be willing to build for i386-family systems?

I tested these changes and, to the best of my knowledge, backrest and restic both appear to be functioning okay on my i686 computer.

This would mostly be for my convenience. If you don't want to run these builds (to save on CI minutes or whatever), that's fine too—I can maintain a fork of the repo with the updated GoReleaser.

Thanks! :heart:

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -35,6 +35,7 @@ builds:
       - amd64
       - arm64
       - arm
+      - "386"
     goarm:
       - 6
       - 7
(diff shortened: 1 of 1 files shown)


garethgeorge:
Happy to support i386 , running release pipeline to verify that the containers build on the older arch.

pwbriggs:
I'm not familiar enough with the CI to know if there's a way I can test the preview docker image directly, but I built the docker image using the binaries output by the preview job, and they seem to run fine. Thanks!

PR #1208, opened 2026-04-24
refactor: upgrade to goreleaser v2

quagsirus:
## What
This is probably something that should be done, since it's impossible to install goreleaser v1 locally with Go v1.25+
I'm pretty new to the Go ecosystem though so please treat this more as a "getting started on the migration" rather than a fully tested complete solution

## Breaking
This would result in the `-<platform>` getting dropped from the docker image names and probably break some people's compose files :/

## Not done
Updated the `brews` section of `.goreleaser.yaml` to the new [`homebrew_casks`](https://goreleaser.com/resources/deprecations/#brews) format since I have no way to test it on a Mac server


garethgeorge:
Thanks for getting this started -- ran into a few issues locally verifying the changes. The image naming incompatibility is awkward but can be called out in the release notes, generally worth moving to v2.

quagsirus:
Sorry about all the mistakes but glad to have helped somewhat 😅

PR #1209, opened 2026-04-26
fix: show repo URI in sidebar tooltip instead of repo ID

alan-vle:
The tooltip on repo items in the sidebar was showing the repo ID, which is redundant since it's already the visible label. This changes it to show the repo URI (e.g. S3 bucket URL, local path).


garethgeorge:
Thank you for contributing this -- I think perfectly reasonable change. Merging.

PR #1222, opened 2026-05-04
fix: set proper path to allow using rclone from brew install

### LINKED ISSUE #1180: [macOS] Homebrew service fails to find rclone/restic due to missing PATH in plist
When installing backrest via Homebrew on macOS (specifically Apple Silicon), the background service fails to locate rclone (and potentially restic) even if they are installed and available in the user's terminal. This is because the generated launchd service (.plist) does not include the Homebrew binary path in its environment.

**To Reproduce**
Install backrest via homebrew tap on Apple Silicon Mac.
Start the service: brew services start garethgeorge/backrest-tap/backrest.
Open Web UI.
Error: rclone not found in PATH (despite being available in /opt/homebrew/bin/rclone).

Expected behavior
The service should inherit or explicitly define a PATH that includes /opt/homebrew/bin and /usr/local/bin so it can find its dependencies.

Environment:
OS: macOS (Apple Silicon)
Installation Method: Homebrew Tap
Backrest Version: 1.12.1 (and likely earlier)

**Suggested Fix**
The homebrew.mxcl.backrest.plist should include an EnvironmentVariables block. Adding the following to the .plist resolves the issue:

```
XML
<key>EnvironmentVariables</key>
<dict>
  <key>PATH</key>
  <string>/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
</dict>
```

nvtkaszpir:
Fixes #1180
tested when installing backrest 1.13.0

```text
2026-05-04T17:32:57.537+0200	INFO	backrest starting	{"version": "1.13.0", "commit": "a389bbc7e2bd344b9cace3a1bb88f73cb224128f", "log_dir": "/Users/michalsochon/.local/share/backrest/processlogs"}
2026-05-04T17:32:57.631+0200	INFO	restic binary "/opt/homebrew/bin/restic" in $PATH matches required version 0.18.1, it will be used for backrest commands
...
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -462,8 +462,8 @@ create_launchd_plist() {
     <true/>
     <key>EnvironmentVariables</key>
     <dict>
-        <key>PATH</key>
-        <string>/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+        <key>PATH</key>        
+        <string>/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
         <key>BACKREST_PORT</key>
         <string>$BACKREST_PORT</string>
     </dict>


nvtkaszpir:
ping @garethgeorge

garethgeorge:
LGTM, thanks for contributing a fix to this and apologies for the slow review.

PR #1236, opened 2026-05-17
chore: install script update, replace cp and chmod with install command

tredondo:
[`install`](https://unix.stackexchange.com/questions/104982/why-use-install-rather-than-cp-and-mkdir) is specifically designed for this cp + chmod flow.


garethgeorge:
LGTM, this is a good refactor. Thanks for contributing.

Issue #978, opened 2025-11-01
SQlite3 error on 1.10.0


florianmski:
**Describe the bug**
After updating from 1.9.2 to latest 1.10.1 I'm getting the following logs, backrest just won't start and loop on these same 4 lines.

```
2025-11-01T13:37:27.939+0100    INFO    backrest version 1.10.1@b9181dc00d7c595c0ff1eae333c6ecdecee96b74, using log directory: /data/processlogs
2025-11-01T13:37:28.194+0100    INFO    restic binary "/bin/restic" in $PATH matches required version 0.18.1, it will be used for backrest commands
2025-11-01T13:37:33.269+0100    WARN    operation log may be corrupted, if errors recur delete the file "/data/oplog.sqlite" and restart. Your backups stored in your repos are safe.
2025-11-01T13:37:33.269+0100    FATAL    error creating oplog: run multiline query: sqlite3: disk I/O error
```

I've tried to follow the warning instructions and delete oplog.sqlite but I'm still experiencing the same, the file is recreated but stays blank.

**Platform Info**

- OS and Architecture: Linux 3.10.108 x86_64 GNU/Linux synology_braswell_716+
- Backrest Version: 1.10.0 and 1.10.1

**Additional context**
Looking at the changes that went into 10.0.0 it seems another [library](https://github.com/garethgeorge/backrest/compare/v1.9.2...v1.10.0#diff-21162f07a37719dece8ce6980ad14beed8c409d2c7e0786878a2ba0598e95f8cL39) to write the sql is being used, maybe that's where the problem comes from? 
I'm running backrest on a quite old Synology and so far it has been working great. I would be tempted to say it's a problem with my docker / filesystem setup but it does work well with other containers and also wouldn't explain why it works great on 1.9.2 but not 1.10.1.
Moving back to 1.9.2 works again so I've got a fix for now at least.
Thanks for the great software and let me know if I can help pin down the issue further.

garethgeorge:
Sorry you're running into this & thanks for the detailed report / steps you've taken to root cause the versions involved.

It definitely could be the case that the sqlite driver is now incompatible with certain filesystems / OSes which isn't great. Interested to track down what exactly you're running into -- are there any details about the filesystem you're running on you can share? is it ext4?

florianmski:
The filesystem I'm using is btrfs

garethgeorge:
[...]

PR #1248, opened 2026-06-01
Added a container build using sqlite3_flock

hanjo:
Added a build option using sqlite3_flock to get a working container for linux servers running older kernel versions as reported in https://github.com/garethgeorge/backrest/issues/978

This needs review. I believe this should work, but I'm not an experienced developer, so it might not 🙃

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -43,6 +43,26 @@ builds:
       - 7
     ldflags:
       - -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}
+  - id: linux-flock
+    main: ./cmd/backrest
+    binary: backrest
+    env:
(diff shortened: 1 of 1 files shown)


garethgeorge:
LGTM, thanks for contributing this -- I think putting it in a separate image is a reasonable way to go here.  Running CICD workflows for it.

hanjo:
Thanks @garethgeorge, appreciate you looking into this.

Not sure why the tests are failing. Is this even related to `sqlite3_flock`?
```
ERROR	sync handler stream error: syncerror CONNECTION_STATE_ERROR_AUTH: pairing failed: unknown peer and no pairing secret provided
```
Looks more like an issue in the testing pipeline.

And for windows: the change should only add another linux image, so I guess there should be no impact to the windows test...

Let me know if I can do something about those errors.

Issue #1257, opened 2026-06-12
Individual Repo Stats not reflected in Dashboard


Yeraze:
I have 6 systems all backing up to my restic server, 6 different repos.  I have backrest set up to give me some stats. I'm not letting backrest drive the actual backups, that's all being done via cron and classic configuration work where each node drives it's own backup.

I was able to setup the daily check and prune operations so that each source shows data for the Stats graphs, which I love!  However, the main dashboard shows straight zeros all the way down after 2 weeks of operation.

How do I get the main dashboard to give me actual data of backup sizes and such?

Dashboard:
<img width="995" height="1043" alt="Screenshot 2026-06-11 at 8 33 12 AM" src="https://github.com/user-attachments/assets/d7622524-b484-4164-9185-af58ebf168d9" />

One of the repo Stats pages:
<img width="1853" height="967" alt="Screenshot 2026-06-11 at 8 34 21 AM" src="https://github.com/user-attachments/assets/8ec327e9-f34c-4cfe-8e4e-b38e663f97d8" />

_Originally posted by @Yeraze in https://github.com/garethgeorge/backrest/discussions/1252_

garethgeorge:
Afraid this isn't supported on the home dashboard page (you aren't missing it :smile:) this is only a feature of the repo stats tab view at the moment. But can definitely be tracked as a feature request.

connexsensei:
I may have something similar yet a bit different but might be related : 

I setup backrest on my server.
I setup backrest on a client.
I'm using docker for both.
Repo is on the server.
Client backup to the server through sftp.

The client and the server are paired, but I didn't shared the repo, and didn't gave permission to receive shared repo on the client : the repo path are (obviously) completly different on the server (local path, since the repo is on the server) and on the client (using sftp to connect and backup to the server).

The Dashboard on the client shows proper backup status and information, as well as the stats, but NOT on the server (about the client backups, and although the repo is on the server).

The server can see the various backups done, can see the snapshots, I can even browse the backup files from the server (I set the server and the client with the same password on the repo), but the stats won't show on the server Dashboard page for that repo nor in the stats view.

I indexed the snapshot, I ran a check to no avail.

Is this similar in a sens that since everything is initialted on the client side and there is no push or shared repo thing ?

jogotcha:
[...]

PR #1258, opened 2026-06-13
feat(webui): reassurance-first redesign of the summary dashboard

jgalea:
Hi @garethgeorge, the recent Chakra overhaul made this clean to build on, thank you for that. I put together a reassurance-first redesign of the Summary view and wanted to run it by you rather than assume it fits. Very happy to adjust the direction, keep or restore anything, or scope it down, your call on whether this is where you want the UI to go.

---

This PR redesigns the Summary dashboard around a reassurance-first model. A hero verdict answers "are my backups healthy?" at a glance, each plan gets a card that leads with plain-language status and a day-by-day 30-day history strip in place of the per-run bar chart, and the card surfaces next run, destination, protected size, last upload, and retention in plain words. A recent-activity timeline across plans sits below. The MultihostSummary and System Info sections are unchanged; repos render a slimmed card.

It is built with Chakra theme tokens so light and dark both work, framer-motion for the running state, and the existing Connect client. New user-facing strings are added to the paraglide messages (English values, copied into the other locale files as a fallback for now). `tsc --noEmit` and the production build pass.

## Screenshots

Light:

Dashboard, light mode

Dark:

Dashboard, dark mode

_(Screenshots use demo data.)_

## Notes for review

- Each plan card issues a `getOperations(lastN)` call to build the history strip and the protected-size figure. For many plans this could sit behind a dedicated batch endpoint; happy to do that if you'd prefer.
- Repo cards are slimmed (no schedule or history strip) since repo summaries don't carry per-day data. Can extend if wanted.
- `StatsPanel` is no longer referenced by the dashboard after this change. I left it in place rather than delete it here; remove at your discretion.
- All hero, card, activity, and history-strip copy is i18n-extracted. The non-English locales currently hold the English string as a fallback, happy to run the translation pipeline for them.


jgalea:
Hi @garethgeorge, no rush on this. I know the summary UI is recent work of yours (#1034), and a full redesign is a lot to land as a single PR. If it's easier to react to the idea before the code, I'm glad to move it to a Discussion, or to scope it down to just the hero verdict and the 30-day history strip and leave the rest of your layout intact. Whichever is least disruptive for you. If there's a direction you'd prefer, tell me and I'll adjust.

Yeraze:
Would love to see this integrated and released.. Would this resolve my issue of https://github.com/garethgeorge/backrest/issues/1257 ?

garethgeorge:
Thank you for the contribution, downloaded this to try it out locally and generally it looks great. Apologies for the late review.

I think I'm happy to merge this as is, I'm going to apply some followups to it to bring back the optimized endpoint for fetching the dashboard, appreciate you calling out the perf issue -- I can certainly help with that.

I see a few unlocalized strings that I can take a pass at and also happy to kick off the translation pipeline.

The only real regression I see that I'd love to tackle in a UI followup, or if you're interested and have bandwidth is: hover details on the colored blobs for each backup, the prior UI would show stats for the run e.g. duration etc when hovered.

PR #1259, opened 2026-06-13
feat: Add /healthz and /readyz endpoint for liveness and readiness probes 

### LINKED ISSUE #1235: Add an unauthenticated /health endpoint for liveness monitoring
### Is your feature request related to a problem? Please describe.

I run Backrest behind a reverse proxy (Caddy) with forward-auth (Authelia) in front of it, and I'd like to monitor it with Uptime Kuma. There is currently no way to do a simple liveness check:

- All paths require authentication once forward-auth is in front, so the monitor can't reach Backrest without being logged in.
- The only path the Backrest server itself answers with `200` is `/`, which serves the full SPA shell — too broad to whitelist in an auth-bypass rule.
- The static assets are content-hashed (e.g. `favicon-KELWFrrG.png`, `index-B5WLpSEK.js`), so they change on every release and can't be pinned as a stable bypass path.
- There is no `/health`, `/healthz`, `/v1/health` or similar — all return `404`.

This makes it impossible to set up a robust health check without either exposing the whole app or pinning an asset path that breaks on the next update.

### Describe the solution you'd like

A dedicated, unauthenticated endpoint such as `GET /health` (or `/healthz`) that returns `200 OK` with a small static body whenever the server process is up. It should:

- not require authentication, so it works behind an authenticating reverse proxy and can be whitelisted as a single stable path,
- be a stable, non-hashed URL that doesn't change between releases,
- return only liveness info (no repo/snapshot data).

This is the same pattern many self-hosted apps already provide and would make Backrest easy to monitor with Uptime Kuma, Gatus, Docker healthchecks, k8s probes, etc.

### Additional context

Tested on a current Backrest release behind Caddy:

/             -> 200 OK
/favicon.ico  -> 404 Not Found
/healthz      -> 404 Not Found
/health       -> 404 Not Found
/v1/health    -> 404 Not Found

Related upstream discussion in restic/rest-server: https://github.com/restic/rest-server/issues/147

lollo03:
This change introduces new unauthenticated `GET /healthz` and `GET /readyz` endpoints. 

These are primarily intended for use as liveness and readiness probes in environments like Kubernetes.

Changes:
* Adds handlers in `internal/api/health/health.go`.
* Registers the `/healthz` and `/readyz` routes in the main application router.
* Includes unit tests to verify the handlers' behavior.
* Updates `api.md` to document the new endpoints.

The `/healthz` endpoint simply checks if the API server is alive, while `/readyz` the database is accessible.

Fixes #1235

This is a proof of concept; I am seeking input on a more robust implementation.


nvtkaszpir:
with auth required endpoint you could just check if the metrics endpoint returns 401 :D

lollo03:
@nvtkaszpir You're right, though that approach isn't quite standard. I've also added a /readyz endpoint, which should be more useful here. Any feedback is welcome!

internal/api/health/health.go
```diff
@@ -1,13 +1,61 @@
...
+			w.WriteHeader(http.StatusServiceUnavailable) // 503
+			json.NewEncoder(w).Encode(ReadyResponse{
+				Status: "DOWN",
+				Reason: "Configuration not loaded or invalid",
+			})
+			return
+		}
+
+		// 2. Check if database is available with a timeout
+		ctx, cancel := context.WithTimeout(r.Context(), 1*time.Second)
```
nvtkaszpir: 1 second timeout for sqlite is way too low, I would recommned 10s

lollo03: Done, thank you!

lollo03:
Hello @garethgeorge can you review this PR? This could be useful

garethgeorge:
Hey -- apologies for the late review. Thanks for the interest in contributing! Generally, adding adding a health check and readiness probe endpoint sounds good to me, but a few thoughts on the current impl

 1. sharedKvdb is the kv/peer-state DB, but isn't the the operations log (opLog/sqlitestore) which is the more important one. Probably want the healthcheck to check the oplog at a minimum, it may be worth just having the healthcheck actually query the operation log with a short limit to test that it can get well formatted results.
  2. 10s is a long timeout, set this to e.g. 1s.
  3. configMgr.Get() failing is unusual -- typically this can only happen on startup and a failure blocks the binary from becoming healthy.

lollo03:
Hello, implemented your suggestions. Let me know what do you think.

lollo03:
@garethgeorge would you please review the changes?

internal/api/health/health.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"encoding/json"
+	"net/http"
+	"time"
+
+	v1 "github.com/garethgeorge/backrest/gen/go/v1"
+	"github.com/garethgeorge/backrest/internal/oplog"
+)
+
+const readyTimeout = 2 * time.Second
```
nvtkaszpir, line 12: hardcoding ready time? I'd give it a bit more such as up to 10s  (which is a limit for sqlite transactions) or allow to adjust it via config. Just a suggestion.

nvtkaszpir (approved):

internal/api/health/health.go
lollo03: @nvtkaszpir  Added the BACKREST_READY_TIMEOUT env var and cli option for configuring the timeout, also the default now is 10 seconds

PR #1265, opened 2026-06-28
fix(webui): retry initial config load before failing to the error screen

jgalea:
This PR makes the initial config load tolerant of transient failures instead of dropping the whole UI to the "Failed to load configuration" screen on the first hiccup.

AuthenticationBoundary raced getConfig against a single 5s timeout and ran once. If that one attempt timed out or hit a transient error (backend busy during a backup, a backgrounded tab throttling timers, a brief connection stall), the app fell through to the error state and the only way back was a full page reload. I hit this reliably while a large backup was running: the backend answered in under a millisecond, but the lone config request still missed its window once and the dashboard locked itself out.

The change retries transient failures (timeout, Unavailable, DeadlineExceeded, or a code-less error) up to four times with exponential backoff and a 10s per-attempt timeout before surfacing the error. Unauthenticated still goes straight to the login modal, and a genuine non-transient error still fails fast without pointless retries. The Retry button now re-runs the loader in place via useCallback instead of reloading the page.

The happy path is unchanged; this only affects what happens when that first request fails.

### DIFF
--- a/webui/src/app/App.tsx
+++ b/webui/src/app/App.tsx
@@ -1,4 +1,4 @@
-import React, { Suspense, useEffect, useMemo, useState } from "react";
+import React, { Suspense, useCallback, useEffect, useMemo, useState } from "react";
 import {
   FiCalendar,
   FiDatabase,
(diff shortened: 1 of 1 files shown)


garethgeorge:
LGTM, seems like a nice reliability improvement. Thanks for contributing this.

PR #1285, opened 2026-07-07
fix: scheduled forget never runs for repos that only receive external snapshots

dobbry-vechur:
If a repo only receives snapshots created outside backrest, the scheduled forget never becomes due. My setup: rest-server on the backrest machine, laptop pushes restic backups to it, backrest handles retention with a repo-level forget policy.

External snapshots never produce an `OperationBackup`. The indexer records them as `OperationIndexSnapshot`. The `foundBackup` scan in `Next` only looked at `OperationBackup`, so an external-only repo always looks idle. With `CLOCK_LAST_RUN_TIME` that resets the schedule base to now when it sees no backups, so restarts and config changes kept pushing the run out.

Fix: count a live (`Forgot == false`) `OperationIndexSnapshot` as activity. Index ops are timestamped from the snapshot's own time, so they slot into the scan the same way backup ops do.

Also dropped a default instance-ID filter that `testTaskRunner.QueryOperations` was adding but prod `taskRunnerImpl` doesn't. It would have hidden unassociated ops from the regression test, since `Next`'s query is deliberately instance-unscoped. Neutral for existing tests, they all seed ops with the config instance ID.

Tests: four scheduled-forget cases in `TestScheduling`. The indexed-snapshot case fails on main.

### DIFF
--- a/internal/orchestrator/tasks/scheduling_test.go
+++ b/internal/orchestrator/tasks/scheduling_test.go
@@ -65,6 +65,14 @@ func TestScheduling(t *testing.T) {
 						Clock: v1.Schedule_CLOCK_LAST_RUN_TIME,
 					},
 				},
+				ForgetPolicy: &v1.ForgetPolicy{
+					Schedule: &v1.Schedule{
+						Schedule: &v1.Schedule_MaxFrequencyHours{
+							MaxFrequencyHours: 1,
+						},
+						Clock: v1.Schedule_CLOCK_LAST_RUN_TIME,
+					},
+				},
 			},
 		},
 		Plans: []*v1.Plan{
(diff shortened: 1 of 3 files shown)


dobbry-vechur:
@garethgeorge sorry to bother, but wdyt?

garethgeorge:
Hi, appreciate the interest in contributing a fix here! Sorry for the late reply -- I've gone ahead and removed the shouldSkip behavior which is yielding merge conflicts, but I'm happy to merge the change to the Next() defn with merge conflicts resolved

PixelJonas:
I've just hit this issue - @dobbry-vechur anything I can do to support resolving the merge-conflicts?

dobbry-vechur:
Hey @PixelJonas I'll make sure to update this PR today.

dobbry-vechur:
@garethgeorge - updated, resolved conflicts, should be ready for merge.

PR #1294, opened 2026-07-12
fix: exit after printing version

phanirithvij:
In tray mode `backrest -tray -version` prints version and app stays open in the tray.

With this pr it will exit properly.


garethgeorge:
Thank you, inconsistent behavior. Fix LGTM.

PR #1295, opened 2026-07-12
fix: quit tray on graceful shutdown

phanirithvij:
Currently `ctrl+c` when backrest is run via `-tray` doesn't quit the tray.


garethgeorge:
Good catch, thanks for PRing the fix.

PR #1308, opened 2026-07-19
fix: preserve indexed snapshots in dashboard history

### LINKED ISSUE #1307: Dashboard missing snapshots for one repo
**Describe the bug**
The new dashboard stopped showing older successful snapshots for one of my repos. This affects all plans on my sftp repo, and none on my external disk repo.

To be clear, this is a display bug. These snapshots haven't been forgotten or pruned, they still exist within restic. In fact, one of the affected plans intentionally doesn't have a retention policy set, so if they had been forgotten, something would be seriously wrong.

Failed snapshots for these days are still displayed, but successful ones (and presumably ones with warnings) are just mysteriously missing. Days where all snapshots have gone missing are displayed greyed out with a hover text of "Backups up to date as of this day".

**To Reproduce**
Not sure what happened here, I didn't do anything special

**Expected behavior**
Backrest should display all snapshots

**Screenshots**
Successful snapshots are still listed for yesterday:

Image

But missing for the previous days:

Image

Image

Even though they still exist in restic:

Image

**Platform Info**

- OS and Architecture: Darwin arm64 (M3 MacBook Pro)
- Backrest Version: backrest version: 1.14.1, commit: 875c9cb51e626a75276a64da42c6a9ec0695010a (via brew)

**Additional context**
Not the same as #1304, as my snapshots haven't been forgotten and aren't listed as 'missed'.

As a side node, do we _need_ to colour the whole day red if, say, two backups failed due to harmless sleep mode or connection drops and 101 succeeded? 😆 Seems a bit silly, I'd love a toggle to configure that.

floze-the-genius:
Fixes #1307.

## Summary
- use live indexed snapshots as fallback evidence for dashboard history when the matching backup operation is no longer present
- deduplicate fallback snapshots against backup operations by snapshot ID
- exclude forgotten snapshots while preserving per-day byte details and overdue tracking

## Tests
- `go test ./internal/api -count=1`
- `go test ./internal/orchestrator/tasks -count=1`
- `go vet ./internal/api`
- `go test ./... -count=1` (with the CI-style fake `webui/dist/index.html.gz` embed)

### DIFF
--- a/internal/api/summarydashboard.go
+++ b/internal/api/summarydashboard.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"slices"
 	"time"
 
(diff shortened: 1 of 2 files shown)


garethgeorge:
Hi, thanks for the interest in contributing -- would you mind including some screenshots / explaining a in a bit more detail the problem you're seeing and how this CL is addressing it?

garethgeorge:
Okay looking deeper -- seeing a few lingering issues:

  - The history strip gets fixed but the counters next to it don't -- the fallback never updates backupsSuccess30, backupsExamined, the byte totals, protectedBytes, or backupChart. Those all render on the same card (SummaryDashboard.tsx:596-623, and :179-190 where the status dot is derived from recentBackups), so in the case this is meant to fix you'd see green days sitting above "0 ok" with an idle dot. 
  - Both new maps hold every snapshot in the oplog, and twice over since each one goes into the plan acc -- this is quite a bit of memory and needs to be avoided.  The start time of a snapshot will always be after the time the backup operation ran, and will be somewhat close, we should be able to do something more clever here.
  - nit: applyIndexedSnapshotFallbacks mutates state from inside finalize, so calling finalize twice would double count. Probably cleaner to call it once at the end of the walk.

floze-the-genius:
Updated the implementation in `f7b7704` to cover the remaining dashboard state:

- indexed-snapshot fallbacks now update success/examined counts, byte totals and averages, `protectedBytes`, daily history, staleness evidence, and `backupChart` through the same accounting path;
- the newest-first oplog walk retains at most one unmatched indexed snapshot per plan instead of maps containing every snapshot;
- pending fallbacks are emitted in newest-first order;
- fallback application now runs once after the oplog walk, outside `finalize`, so repeated finalization does not double-count.

Added coverage for counters, bytes, chart entries, deduplication, forgotten snapshots, ordering, and repeated finalization. Validation: `go test ./internal/api -count=1`, `go test ./internal/orchestrator/tasks -count=1`, and `go vet ./internal/api`.

floze-the-genius:
Controlled local before/after reproduction, using the same config and operation log for both binaries. The fixture contains three successful `OperationIndexSnapshot` records (4, 2, and 0 days ago) and no `OperationBackup` records.

**Base `148ba12c`:** the dashboard ignores those live indexed snapshots and renders the empty state.

Base dashboard: Waiting for first backup

**PR head `f7b7704a`:** the same data renders `Everything is backed up`, three successful history days, `Protected 300 MiB`, and `Last upload 30 MiB`.

PR dashboard: indexed snapshot history restored

This verifies that the fallback now feeds the complete dashboard accounting path rather than only coloring the history strip. Validation on the PR head: `go test ./internal/api` and `pnpm run build`.

floze-the-genius:
Hi @garethgeorge — following up on the requested changes: `f7b7704` addresses the dashboard counters, chart, and staleness path; bounds indexed-snapshot retention; and moves fallback application out of `finalize`. The controlled before/after reproduction is above, and the branch is still current with `main`. Is there anything else you would like changed?

Issue #1314, opened 2026-07-21
"You can’t use this version of the application “Backrest.app” with this version of macOS." (macOS 15.7.7)


tbodt:
Using backrest 1.14.1 download from releases

<p><img width="372" height="358" alt="Image" src="https://github.com/user-attachments/assets/a63694c6-59ad-4fcd-b877-a1a563d56dbf" /></p>

```
$ sw_vers
ProductName:		macOS
ProductVersion:		15.7.7
BuildVersion:		24G720
```

tbodt:
Go's native linker sets the minimum macos version in the binary to the minimum version go actually supports (for go 1.26, macos 12). But cgo doesn't know how to set this, so it uses xcode's default of the major version that the compiler is running on. The best way I found to fix this is to set environment variables `CGO_CFLAGS="-mmacosx-version-min=12" CGO_LDFLAGS="-mmacosx-version-min=12"` during the mac build. (This doesn't affect command line usage since the kernel doesn't actually check this version, it only breaks launching it as a desktop app.)

PR #1316, opened 2026-07-21
Build mac binary with correct minimum macos version

tbodt:
#1314


garethgeorge:
Thanks for contributing this fix! LGTM

PR #1317, opened 2026-07-22
fix: put status text in disabled menu item on macos

tbodt:
Tooltips on menu bar icons are unexpected on macos, the idiomatic solution is disabled menu items.


garethgeorge:
LGTM -- thanks for contributing this.

Issue #919, opened 2025-09-26
notify-send hook command fails on Ubuntu


tredondo:
**Describe the bug**
Attempting to run `notify-send` as a hook command on Ubuntu 25.04 fails with an error spawning the "dbus-launch" command line.

**To Reproduce**

1. Add a hook on CONDITION_SNAPSHOT_END
2. For 'Script' enter `DISPLAY=:1 notify-send -u critical "Backrest update" {{ .ShellEscape .Task }}`
3. Error behavior ON_ERROR_IGNORE
4. Submit
5. Wait for the plan to run

**Expected behavior**
`notify-send` should send the notification

**Actual logs for operation Run Hook**
```
[script] [sh]
[script] DISPLAY=:1 notify-send -u critical "Backrest update" 'backup for plan "B2-stuff"'
[output] Error spawning command line “dbus-launch --autolaunch=2c97b974a1c94e8285a7049030e62dd5 --binary-syntax --close-stderr”: Child process exited with code 1
[tasklog] 00:06:07.992Z	error	run hook command hook plan/B2-stuff/hook/0	task failed	{"error": "exit status 1", "duration": 0.01717468}
```

**Platform Info**

- OS and Architecture: Ubuntu 25.04 amd64, KDE 6
- Backrest Version 1.9.2

**Additional context**
First I entered the command without `DISPLAY=:1` but I got an error "cannot autolaunch d-bus without x11 $display"

garethgeorge:
Hi, I think this is more of a scripting issue than a platform bug. What user is backrest running as / is it running in a context where they'd have access to the DISPLAY environment variable for your logged in user session?

tredondo:
It's running as my own user (not root or anything else). I have access to $DISPLAY in a Konsole terminal:

```
$ echo $DISPLAY
:1
```

kashyl:
I have the same issue. Would be really nice to get notifications working on KDE without the need of a user-level bridge or similar workaround.

```
[script] [sh]
[script] notify-send -a "Backrest" "test - test" "Backup starting..."
[output] Failed to show notification: Cannot autolaunch D-Bus without X11 $DISPLAY
[tasklog] 17:43:23.216Z	error	run hook command hook plan/test/hook/0	task failed	{"error": "exit status 1", "duration": 0.007717461}
```

Issue #868, opened 2025-08-13
Help to send notifications contents to a webhook


lordraiden:
I need help creating the script to send the ouput of the notifications to a webhook. I am getting parsing errors....
Can someone help me?

/usr/bin/curl -X POST \
  -H "Content-Type: application/json" \
  -H "X-N8N-TOKEN-NOTIFICATION: aaa" \
  --data-binary '{"summary": "{{ .Summary }}"}' \
  "https://n8n.test.es/webhook/f70542ed-645e-42f6-91e3-18606b9c6257"

<img width="1063" height="673" alt="Image" src="https://github.com/user-attachments/assets/11f16a73-2355-4c95-ac9a-299e6316b67f" />

This is the output

[script] [sh]
[script] /usr/bin/curl -X POST \
[script]   -H "Content-Type: application/json" \
[script]   -H "X-N8N-TOKEN-NOTIFICATION: aaa" \
[script]   --data-binary '{"summary": "
[script] Backrest Notification
[script] Task: prune repo "Local" at 2025-08-13T17:50:28+02:00
[script] Event: prune success
[script] 
[script] "}' \
[script]   "https://n8n.test.es/webhook/f70542ed-645e-42f6-91e3-18606b9c6257"
[output]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[output]                                  Dload  Upload   Total   Spent    Left  Speed
[output] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   258  100   144  100   114   3803   3011 --:--:-- --:--:-- --:--:--  6972
[output] {"code":422,"message":"Failed to parse request body","hint":"Bad control character in string literal in JSON at position 13 (line 1 column 14)"}[tasklog] 17:50:28.597Z	info	run hook command hook repo/Local/hook/1	task finished	{"duration": 0.046643806}

lesleyxyz:
Use Shoutrrr instead using the "generic webhook service":
https://containrrr.dev/shoutrrr/v0.8/services/generic/

PR #1323, opened 2026-07-24
feat: add test hook button for notification hooks

caco3:
Adds a **Test** button to every hook form so users can immediately verify a hook (e.g. Telegram, Discord, Slack) with a test message before saving:

image

## What's changed
- Adds a `TestHook` RPC to the backend.
- `TestHook` executes the provided hook with synthetic `HookVars` and returns an error if the hook fails.
- `TestHook` uses a real `TaskRunner` via `orchestrator.NewReadOnlyTaskRunner`. A `readOnlyTaskRunner` decorator wraps the real runner and no-ops state-mutating methods (`CreateOperation`, `UpdateOperation`, `DeleteOperation`, `ScheduleTask`, `LogrefWriter`) so the test cannot create operations, schedule tasks, or write log refs.
- Adds a play button to each hook form in the UI that serializes the current hook, calls `testHook`, and shows a success or error toast.

## Technical details

- `proto/v1/service.proto`: adds the `TestHook` RPC.
- `gen/go/v1/service.*.go`: generated client/server code for the new RPC.
- `internal/orchestrator/taskrunnerimpl.go`: adds `readOnlyTaskRunner` and the `NewReadOnlyTaskRunner(orchestrator, task)` constructor.
- `internal/api/backresthandler.go`: implements `TestHook` by creating a minimal `tasks.GenericOneoffTask` and running the hook through `orchestrator.NewReadOnlyTaskRunner`.

## Notes

- This is mainly AI driven; my Go skills are very limited, but I reviewed and tested the Telegram hook.
- I only tested the Telegram hook as I have no infrastructure for the other services.


caco3:
There are many open issues which will benefit from this feature:
- https://github.com/garethgeorge/backrest/issues/919
- https://github.com/garethgeorge/backrest/issues/868

internal/api/backresthandler.go
```diff
@@ -883,6 +885,85 @@ func (s *BackrestHandler) GeneratePairingToken(ctx context.Context, req *connect
...
+		Task:     "test hook",
+		Event:    event,
+		Repo:     &v1.Repo{Id: "test-repo", Guid: "test-guid"},
+		Plan:     &v1.Plan{Id: "test-plan", Repo: "test-repo"},
+		CurTime:  time.Now(),
+		Duration: 0,
+	}
+
+	ctx = orchestratorlogging.ContextWithWriter(ctx, io.Discard)
+	if err := handler.Execute(ctx, hook, vars, &testHookRunner{}, event); err != nil {
```
garethgeorge: I think it'd be fine to use a real [taskrunnerimpl](https://github.com/garethgeorge/backrest/blob/main/internal/orchestrator/taskrunnerimpl.go] here.

Let's introduce a ReadOnlyTaskRunnerDecorator in that same file which wraps a task runner and no-ops methods that mutate state

 1. CreateOperation
 2. DeleteOperation
 3. ScheduleTask
 4. LogrefWriter

wdyt?

garethgeorge:
Thanks for the interest in contributing this -- agree this is much wanted and its a great contribution!

I think it'd actually be reasonable to inject a real but decorated taskrunnerimpl here and decorate it.

caco3:
@garethgeorge Good call — using a real task runner is much cleaner. I added the `ReadOnlyTaskRunnerDecorator` and switched `TestHook` to use it.

### What changed

#### `internal/orchestrator/taskrunnerimpl.go`
- Added `readOnlyTaskRunner`, a decorator that wraps a real `taskRunnerImpl`.
- It overrides the state-mutating methods to no-op:
  - `CreateOperation`
  - `UpdateOperation`
  - `DeleteOperation`
  - `ScheduleTask`
  - `LogrefWriter`
- Every other method (`Config`, `InstanceID`, `GetRepo`, `GetPlan`, `Logger`, etc.) is delegated to the real runner.
- Added `NewReadOnlyTaskRunner(orchestrator, task)` as a helper constructor.

#### `internal/api/backresthandler.go`
- Replaced the hand-written `testHookRunner` stub with a real read-only runner:
  ```go
  testTask := &tasks.GenericOneoffTask{
      BaseTask: tasks.BaseTask{
          TaskName: "test hook",
          TaskType: "test",
      },
      Do: func(ctx context.Context, st tasks.ScheduledTask, runner tasks.TaskRunner) error { return nil },
  }
  runner := orchestrator.NewReadOnlyTaskRunner(s.orchestrator, testTask)
  ```
- Removed the `testHookRunner` and `nopWriteCloser` stub types.

Issue #789, opened 2025-05-26
Reusable Hook Destinations


olivenjoyer:
I'm creating 10+ jobs and repos. In the Hooks section for each one, I have to manually select Discord and then paste in the webhook URL and the text template each time. I must manually go through each one to update the webhook URL later if I want to change it. Very tedious.

**Describe the solution you'd like**
The ability to create a Hook destination once and choose it through a dropdown, Similar to how the repository field is selected when creating a new plan.

Having a shared credential/variable for a hook target would also allow for updating the URL in one place (in the Settings dialogue?) and updating it in all the plans where it is used.

caco3:
Please have a look on https://github.com/garethgeorge/backrest/pull/1324 where I added some improvement for this issue. It does not provide reusable hooks, but the config of an existing hook can now easier be re-used.

Issue #595, opened 2024-11-30
Support global hooks


banto6:
Note: if you have a question or want discussion please post in the [discussions area](https://github.com/garethgeorge/backrest/discussions).

**Is your feature request related to a problem? Please describe.**

Currently, implementing the same notification method for each repository or plan can only be configured individually within the corresponding plan or repository.

**Describe the solution you'd like**

Add the functionality to maintain global Hooks, and then reference the corresponding global Hooks in plans or repositories.

**Additional context**

garethgeorge:
The are architectural reasons that this probably isn't something that makes sense for backrest -- though I do agree that backrest needs a good way to share configuration between repos. 

I think some sort of templating framework would be a good UI feature and is the most likely solution to this problem, though unsure when I'll budget time for it as it introduces some definite complexity. In the meantime, editing the config.json and directly copying hooks between repos can be a pretty good way to quickly configure repos in bulk.

caco3:
Please have a look on https://github.com/garethgeorge/backrest/pull/1324 where I added some improvement for this issue. It does not provide global hooks, but the config of an existing hook can now easier be re-used.

PR #1324, opened 2026-07-24
Autocomplete hook token/ID/URL fields from existing config

caco3:
Adds autocomplete for hook credential fields (bot tokens, chat IDs, webhook URLs, etc.) based on values already present in the saved config. When a user focuses a supported hook field, a dropdown appears with existing values from other hooks identical to the ones for the `paths`. The dropdown also includes values from other hooks in the same form that have not yet been saved, making it easier to configure multiple similar hooks in one session.

image

This PR is an approach to solve the following issues:
- https://github.com/garethgeorge/backrest/issues/789
- https://github.com/garethgeorge/backrest/issues/595

## Notes
-  This PR is AI driven. My Go skills are very limited. But I reviewed it and tested it.
- The implementation avoids the project's combobox component for hook fields because its state machine conflicts with the parent form's per-keystroke re-renders, which caused focus loss. Instead, a plain controlled input with a simple absolutely-positioned suggestion list is used.


garethgeorge:
Hi, thanks for the interest in contributing! 

Afraid for this one I'm concerned about the maintenance cost of serverside autocomplete for these fields. I think a client side only solution would be more interesting-- but I'm also wondering if this is broadly the wrong way of approaching the problem.

I think I'd like to tackle this more comprehensively at some point with a template / inheritance system allowing users to either copy and inherit configuration from other plans-- but some real thinking about the semantics of this are needed.

caco3:
Thanks for the feedback!

I am not sure what you mean with " template / inheritance system". Could you elaborate a bit more, then I can give it a try.

As for the client side focus, how would you want to do this? At least once we would have to read the config file to extract the current values.

caco3:
@garethgeorge  ping :)

PR #1345, opened 2026-08-13
feat: add Pushover.net notification hook

copethomas:
This feature adds support for [pushover.net](https://pushover.net/api) notifications as a new hook type.

Note: This feature has been AI-generated; however, I have experience in these languages. I have reviewed and tested the code locally. My local machine has a newer version of the proto compiler, which has resulted in a bump in the generated files. I can downgrade locally if you wish to make the PR less noisy. 

Let me know what you think!


copethomas:
@garethgeorge Any thoughts on this?

PR #1384, opened 2026-09-14
docs: add Easypanel deployment option

Ahson-Shaikh:
Adds an Easypanel section alongside the Docker Compose section. Easypanel has a one-click deployment template for Backrest: https://easypanel.io/templates/backrest


edvwib:
It sounds weird calling it an "official template" in this repos readme. It sounds like it's the Backrest developer that has created it, which it is not.

This also just feels like an ad.

Ahson-Shaikh:
Hi @edvwib - The intention was to endorse it from Easypanel's side as we own the template for the users deploying it. Although its not an Ad but rather a snippet for integration of one of the deployment providers that has the template listed. Would you like me to move further in this and do some changes? I would really appreciate your suggestion. Thanks!

Ahson-Shaikh:
Hi @edvwib - fair point, and thanks for the patience. I reworded it: it now reads "Easypanel is a self-hosted Docker deployment platform. It has a one-click Easypanel template for Backrest" - so "one-click" and "template" both clearly belong to Easypanel, not something Backrest ships or endorses. Also swapped the plain link for the actual Easypanel deploy badge so it reads as a deploy option rather than prose. Let me know if it still feels off and I'll adjust further.

